PR #34, opened 2015-04-29
Update Datatables.php

openvast:
issue:https://github.com/yajra/laravel-datatables-oracle/issues/33

HI, I try to solve this problem, and success。

path:vendor\yajra\laravel-datatables-oracle\src\yajra\Datatables\Datatables.php
in image line 961 ,define a new column in array ....

can you use that and update the project as soon as fast to solve the problem forever.
English is not good,Thank you for your patience to read^.^^.^.^^.^

### DIFF
--- a/src/yajra/Datatables/Datatables.php
+++ b/src/yajra/Datatables/Datatables.php
@@ -958,6 +958,7 @@ class Datatables
      */
     public function addColumn($name, $content, $order = false)
     {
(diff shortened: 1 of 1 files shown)


yajra:
should be fixed on v4.3.1. Kindly checkout and test. Thanks!

openvast:
thanks

PR #35, opened 2015-05-05
FIX: request input to call built in capture() to retrieve new instance

florin-hallnet:
Hi Arjay, 

I have discovered an issue when creating a new Request instance by passing in the GET and POST superglobals, the Request method will default to GET and somehow this resulted with `$request->input()` returning an empty array in case of POST requests. This was because the server superglobal was not passed in the Request constructor. Please see [this link](https://github.com/symfony/HttpFoundation/blob/master/Request.php#L1339) related to Request method defaulting to GET if the server array is empty.

This issue lead to the initial datatable being populated with data but pagination or limiting would not work as the the server would always return the same results due to input being empty.

This change will use the Request's built in `capture()` method which will return a fully set up instance and the Datatables' `processData()` will still receive the input alone.

Thank you for this very useful package

### DIFF
--- a/src/yajra/Datatables/Datatables.php
+++ b/src/yajra/Datatables/Datatables.php
@@ -184,7 +184,7 @@ class Datatables
      */
     public function __construct()
     {
-        $request = new Request($_GET, $_POST);
+        $request = Request::capture();
         $this->setData($this->processData($request->input()));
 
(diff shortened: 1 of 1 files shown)


yajra:
Thanks for the contribution!

PR #37, opened 2015-05-07
fixes operator typo. searchable: false does not work otherwise

aliselcuk:


yajra:
Good catch! It was like as far as I remember. Thanks a lot! :+1:

PR #45, opened 2015-05-14
Fix for multiple backtick (`) on column filtering

codewizz:
Multiple backtick symbol is generated when doing filtering by column. 

An example of generated query before fix:
SQLSTATE[42S22]: Column not found: 1054 Unknown column '`email`' in 'where clause' (SQL: select count(*) as aggregate from (select '1' as row_count from `subscriptions` where `user_id` = 1 and ```email``` LIKE %test%) count_row_table)


yajra:
Good catch. Thanks for the contribution!

codewizz:
You are welcome!

PR #56, opened 2015-05-19
Fix missing request data when testing with phpUnit

vertoo:
When I tried test controller methods on get data with datatables I found problems. GET array was empty. Same query in browser works fine.
For testing I use laravel/integrated package. Here is example:
```php
/** @test */
public function get_users()
{
    // because url string is long it's truncated for improving readability
    $this->get('/admin/permissions/data?draw=1&columns%5B0%5D%5Bdata%5D=name')
            ->seeIsJson();
}
```
With my tables this change works fine.

### DIFF
--- a/src/yajra/Datatables/Engine/BaseEngine.php
+++ b/src/yajra/Datatables/Engine/BaseEngine.php
@@ -194,8 +194,7 @@ class BaseEngine
      */
     public function __construct()
     {
-        $request = Request::capture();
-        $this->input = $request->all();
+        $this->input = Input::all();
         $this->getTotalRecords(); // Total records
     }
 


yajra:
Travis is failing because you did not import or `use Input`. On the other hand, `Input` is just a Facade of `Request` so I guess your PR will not make any significant change at all? Anyways, will try to use laravel/integrated and check the test too.

yajra:
I just tried laravel/integrated and it works as expected. Please see [datatables demo app](https://github.com/yajra/laravel-datatables-demo) for a basic test where I used it.

yajra:
@vertoo this should be fixed on v5.2 and have tested it using the demo app. In addition, you may also like the new feature I added where you can use the package via dependency injection or the IOC container. Please verify and close this if fixed. Thanks!

vertoo:
@yajra your latest chages fix previous issues. 

About Input - yes, I forgot add `use Input;` when clone my fork for create pull request. Version 5.2 works fine, this can be closed.

Thanks.

PR #67, opened 2015-06-02
fix a bug with  search value is '0'

xujif:
in php
empty('0') == true
so,when search a field value with '0',this function did not work;


yajra:
good catch! thanks for the contribution...

PR #95, opened 2015-07-01
Prevent unnecessary row count query

codewizz:
When data is loaded and filter isn't applied for any column, two identical row count queries are executed: 
```sql select count(*) as aggregate from (select '1' as row_count from `workers`) count_row_table```

The following fix will make engine to execute second count query (to calculate filtered records) only if filter was applied for at least one column.


yajra:
@codewizz, thanks for this PR. will implement this on next release.

yajra:
PR added on v5.5.0. Thanks!

codewizz:
You are welcome!

PR #98, opened 2015-07-02
Database detection from Eloquent builder fix

sleeping-owl:


yajra:
thanks a lot! released on v5.5.2

PR #119, opened 2015-07-14
Update Datatables.php

Aner-Git:
Change the return string to 'datatables'; 

The previous  return string was  mismatched with the register function in DatatablesServiceProvider.php. 

There is also a need to update the 'Quick Installation,' under the Facade section: 

 'Datatables' => 'yajra\Datatables\Datatables',    //currently, incorrect

 'Datatables' => 'yajra\Datatables\Facades\Datatables',   //proposed

### DIFF
--- a/src/yajra/Datatables/Facade/Datatables.php
+++ b/src/yajra/Datatables/Facade/Datatables.php
@@ -11,6 +11,6 @@ class Datatables extends Facade
      */
     protected static function getFacadeAccessor()
     {
-        return 'Datatables';
+        return 'datatables';
     }
 }


yajra:
I think we also need to rename the folder `Facade` to `Facades`. `src/yajra/Datatables/Facade/Datatables.php` to `src/yajra/Datatables/Facades/Datatables.php`

Aner-Git:
For clarity you can, and for 'Laravel' way probably should.  But, dont need to. 

(And Note that the namespace inside Datatables is  yajra\Datatables\Facades; )

composer seem to be able to find it. (grep composer directory to see mapping)

yajra:
@Aner-Git , thanks for the contribution. I also renamed Facade to Facades for proper convention. Released on v3.6.7

PR #126, opened 2015-07-20
Fixed row variables (like DT_RowClass) was not set

Mevrael:
Updated DataProcessor.php to fix row variable names (returned row array keys, for example, DT_RowClass) was set to numeric key.

### DIFF
--- a/src/yajra/Datatables/Processors/DataProcessor.php
+++ b/src/yajra/Datatables/Processors/DataProcessor.php
@@ -72,12 +72,12 @@ class DataProcessor
             $data  = Helper::convertToArray($row);
             $value = $this->addColumns($data, $row);
             $value = $this->editColumns($value, $row);
-            $value = $this->setupRowVariables($value, $row);
             if ( ! $object) {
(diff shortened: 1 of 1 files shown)


yajra:
@Mevrael, thanks for this PR but I think this may cause some issue when we used a value like `setRowClass('id')` for instance and was removed via `removeColumn('id')`.  Can you please paste some snippets to reproduce this issue?

Mevrael:
@yajra, I see. Here is an example.

yajra/laravel-datatables-oracle       v5.8.2

https://gist.github.com/Mevrael/330c816c1cc806e9eef1

set_row_class was not working at all. When I investigated json I found problem at DataProcessor.php. Probably because of Arr::flatten array key DT_RowClass was renamed to 6.

yajra:
Oic, you are not using an object response `make(true)`. Will check this as soon as I can. Thanks for the info!

yajra:
@Mevrael, patch to fix this issue released on v5.8.6 Thanks!

PR #136, opened 2015-07-30
Only publish package configuration

sangar82:
Add provider to vendor publish command.
Only publish the package configuration for this package. With the other command, we are publishing all the config files


yajra:
thanks @sangar82 !

PR #140, opened 2015-08-06
tableAttributes method on Html Builder that allows to specify additional attributes for <table> tag

vitalijssilins:


yajra:
@vitalijssilins, thank your for this PR. However, we can already pass an attributes when you render the table on `public function table(array $attributes = [])`.  And we should be able to change the `id` attribute for cases when we have multiple DT instances in a page.

On the other hand, I think there is a possible bug on `table` method since `id` attribute is required and overriding the attributes without id may cause package to fail. Thanks for the hint.

vitalijssilins:
@yajra can you please provide me with functional HtmlBuilder code how to add specificall classes to generated "table". Currently I use my fork:
```
->tableAttributes([
    'class' => 'display projects-table table table-striped table-bordered table-hover'
])
```

yajra:
Here's a snippet from my inventory project.

**controller**
```php
    public function index(Datatables $datatables)
    {
        $html = $this->htmlBuilder($datatables);

        return $this->setTitle('Inventory Management')
                    ->setContent('admin.inventories.index', compact('html'));
    }
```

**view**
```php
{!! $html->table(['id' => 'inventory', 'class' => 'table table-hover table-bordered table-condensed']) !!}
```

vitalijssilins:
Thank you!

I was so in backend, that I almost forgot about modifying views. :)


---
Vitalijs Silins
IT consultant
+371 22118877

PR #146, opened 2015-08-14
Compound-key support

hiendv:
Sometimes we need compound keys for relationships. E.g. Performing search by the name of the author (author.name) of a post.

### DIFF
--- a/src/yajra/Datatables/Engines/CollectionEngine.php
+++ b/src/yajra/Datatables/Engines/CollectionEngine.php
@@ -123,15 +123,33 @@ class CollectionEngine extends BaseEngine implements DataTableEngine
                 $keyword = $this->request->keyword();
                 foreach ($this->request->searchableColumnIndex() as $index) {
                     $column = $this->getColumnName($index);
-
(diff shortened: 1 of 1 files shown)


yajra:
@hiendv, thanks for the idea and the PR. I think a better way to implement a compound key search is by using Laravel's array helper. Maybe something like below? 

```php
foreach ($this->request->searchableColumnIndex() as $index) {
    $column = $this->getColumnName($index);
    if ( ! $value = Arr::get($data, $column)) {
        continue;
    }

    if ($this->isCaseInsensitive()) {
        $found[] = Str::contains(Str::lower($value), Str::lower($keyword));
    } else {
        $found[] = Str::contains($value, $keyword);
    }
}
``` 

And I think the code above may also work on multi dimensional array and is more flexible? Would you be able to test the solution like above? Thanks!

yajra:
I just released v5.11.1 which includes support for compound key when filtering and ordering collection data. Thanks for the idea.

hiendv:
@yajra Sorry for the late reply. I did check and it worked ! Thank you for the update. Keep up the good work.

PR #152, opened 2015-08-24
Add support for regular expressions in column search

Fl0ux:
Thanks for this great package !

Here is a little patch to add regular expressions search on columns.


yajra:
thanks @Fl0ux, I think this may cause some issues on other database because I think REGEXP is not standard to all RDBMS. But it's a good one though atm.

Fl0ux:
Indeed it seems to work only with MySQL and SQLite.

yajra:
released on v5.11.2. Thanks!

PR #194, opened 2015-09-20
QueryBuilder: ORDER BY *

hiendv:
If we perform a select("*"), the ORDER BY clause will look like this: `ORDER BY * ASC` which causes a query exception.
Alt Text
The temporary fix is modifying `*` column to `id` column.


yajra:
@hiendv Thanks! Released on v5.11.9

PR #197, opened 2015-09-27
collectionengine: fix sorting of relation columns by using seralize

smb:
serialize seems to work (already does in filtering) for relation columns (flattening the relation to array and accessing the column with dot notation)


yajra:
thanks @smb! Released on v5.11.10

PR #227, opened 2015-10-28
use connection grammar to wrap the row_count alias

donnykurnia:
in firebird, it's important to wrap the alias


yajra:
Nice. Good point on `wrap` function. However, can you please fix the failing test? Thanks!

yajra:
I have fixed the test instead. Released on v5.11.12. Thanks a lot!

donnykurnia:
Thanks @yajra

PR #228, opened 2015-10-29
Update Datatables.php

lisa-fehr:
Switched remaining $columns to $this->columns in doOrdering()


yajra:
@lisa-fehr, thanks! Released on v3.6.10.

PR #230, opened 2015-10-29
Cache results

standaniels:


yajra:
@standaniels, thanks for this but I think we should try and make caching optional/configurable? 

Also, I just remember the caching on Laravel 4.2 version and one problem I encounter is that the results/count does not update when a new record was inserted. When using cache, then we should also add some code to unset it somewhere. Maybe we can try something like:

```php
$datatables->of($model)
  ->cache('datatable.key', 'duration')
  ->make();
```

This way we can unset the cache since we know the key to use and caching will only be triggered when `cache` method was called. This will also help us avoid a possible breaking changes. Thanks!

standaniels:
Thanks for your feedback! You're right, I haven't thought of that because our CMS already clears the cache when modifying models. Making caching optional is also a great idea.

Will be continued... :)

yajra:
Please re-submit PR if possible. Thanks!

curunoir:
Hello, if it interests you I tried to make a custom EloquentEngine for caching queries 👍 
https://github.com/curunoir/rememberabledatatables
 
I inspired myself from Dwight Watson package : https://github.com/dwightwatson/rememberable

Issue #240, opened 2015-11-05
Resolve Fractal Manager from the IoC


prYmea:
Hi,

Is it possible to get an instance of the Fractal manager from the IoC in the `render` method (`BaseEngine.php` line 663) like below:

```
   $fractal = app(\League\Fractal\Manager::class);
```

Reason:
That way I can bind the fractal manager in a custom service provider to set a serializer. At the moment there is no way (correct me if I am wrong) to change the transformer serializer in your package.

Thanks!

yajra:
Thanks for bringing this up. I haven't used much of fractal's features and not yet sure on serializer. Will try to implement this on my free time. Or you can also send a PR if you can, since I think you know Fractal more than I do. ^_^ Thanks!

prYmea:
I have submitted a pull request with the simple changes that I mentioned in my original post.

Using fractal [serializer](http://fractal.thephpleague.com/serializers/) one can change the structure of the transformer's output. 

I created the below custom service provider where I can swap out the serializer to a different one or even a custom one if I wanted.

```
use Illuminate\Support\ServiceProvider;

class FractalServiceProvider extends ServiceProvider
{
    /**
     * Register the application services.
     *
     * @return void
     */
    public function register()
    {
        $this->app->bind(\League\Fractal\Manager::class, function($app)
        {
            $fractal = new \League\Fractal\Manager;

            $serializer = new \League\Fractal\Serializer\ArraySerializer();

            $fractal->setSerializer($serializer);

            return $fractal;
        });

        $this->app->bind(\Dingo\Api\Transformer\Adapter\Fractal::class, function($app)
        {
            $fractal = $app->make(\League\Fractal\Manager::class);

            return new \Dingo\Api\Transformer\Adapter\Fractal($fractal);
        });
    }
}
```

yajra:
Should be fixed on v5.12.0. Thanks!

PR #241, opened 2015-11-06
Resolve Fractal Manager from the IoC

prYmea:


yajra:
Based on your comment #240, we would need to add another provider for the serializer to work? What if we just add a `setSerializer` method on datatables? Maybe a flow something like below.

```php
return Datatables::of($model)
  ->setTransformer(ModelTransformer::class)
  ->setSerializer(ModelSerializer::class)
  ->make(true);
```

Maybe we can also add a default serializer on `datatables.php` config file?

prYmea:
That will be great and the default in the config is an excellent idea 👍🏻

yajra:
I just released v5.12.0 that implements support for fractal serializer. Default serializer is now configurable too via `datatables.php` config file

Usage:

```php
return Datatables::of($model)
  ->setTransformer(ModelTransformer::class)
  ->setSerializer(ModelSerializer::class)
  ->make(true);
```

Config:
```php
    'fractal' => [
        'serializer' => League\Fractal\Serializer\DataArraySerializer::class,
    ],
```

PR #243, opened 2015-11-06
count method needs to check for distinct as well as union

lisa-fehr:
"select distinct '1' as row_count" always made count = 1

### DIFF
--- a/src/yajra/Datatables/Datatables.php
+++ b/src/yajra/Datatables/Datatables.php
@@ -327,9 +327,9 @@ class Datatables
     {
         $query = $this->query;
 
-        // if its a normal query ( no union or distinct ) replace the select with static text to improve performance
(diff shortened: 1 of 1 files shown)


yajra:
@lisa-fehr, thanks for the PR. Can you please include `HAVING` too? This was fixed on the L5 version but I was not able to merge it on L4. Below is the L5 snippet. Thanks!

```php
if ( ! Str::contains(Str::lower($myQuery->toSql()), ['union', 'having', 'distinct'])) ....
```

lisa-fehr:
@yajra I updated it to the snippet. Appears to work for me.

yajra:
@lisa-fehr, this was released on v3.6.11 along with your PR on ordering. Thanks a lot!

PR #245, opened 2015-11-09
sort by a multi-line 'select as' query

lisa-fehr:
Allow it to grab the last 'as' in a multi-line select query (within a select query). Also does one line.

Would solve something complicated like this:
```php
->select('table.begin_at', DB::raw('(SELECT table2.id FROM table as t ...) as file_id'))
```

We'll be on L4 for a while due to a large project, hope you don't mind the requests.

### DIFF
--- a/src/yajra/Datatables/Datatables.php
+++ b/src/yajra/Datatables/Datatables.php
@@ -295,7 +295,7 @@ class Datatables
      */
     private function getColumnName($str)
     {
-        $matches = explode(' as ', $str);
+        $matches = explode(' as ', Str::lower($str));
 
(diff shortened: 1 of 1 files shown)


yajra:
@lisa-fehr, I don't mind these pull requests and I appreciate the help. Anyways, can you please make the script case insensitive? Maybe something like below...

```php
$matches = explode(' as ', Str::lower($str));
```

lisa-fehr:
Good idea. Added Str::lower($str)

yajra:
Released on v3.6.12. Thanks!

PR #249, opened 2015-11-12
Update DataTableAbstract.php

Skintillion:
Add the ability to change columns for print, as well as have them differentiate from the exportColumns. When printing one might want Icons to show, but not when exporting to excel or csv for example.


yajra:
Please send the PR on develop branch. And great idea on the printColumns option. :+1:  Thanks!

PR #250, opened 2015-11-12
Update print.blade.php

Skintillion:
Hopefully I did this in the right branch this time =)

Checks for string and still displays if empty, to maintain column structure.


yajra:
I think you still post on the wrong branch? Please use **develop** branch. Thanks! Will also removed the unnecessary branches. ^_^

Skintillion:
Those files didnt exist on the develop branch  when I tired that. Sorry.

PR #274, opened 2015-12-02
use select" table.column as myColumn" the bug

openvast:
return array_shift($matches);
//            return array_pop($matches);

### DIFF
--- a/src/yajra/Datatables/Engines/BaseEngine.php
+++ b/src/yajra/Datatables/Engines/BaseEngine.php
@@ -258,7 +258,8 @@ abstract class BaseEngine implements DataTableEngine
         $matches = explode(' as ', Str::lower($str));
 
         if ( ! empty($matches)) {
-            return array_pop($matches);
+            return array_shift($matches);
+//            return array_pop($matches);
         } elseif (strpos($str, '.')) {
             $array = explode('.', $str);
 


yajra:
I think this PR will break this fix #245 . Using `array_shift` will return `table.column` instead of the alias `myColumn`. Does using the alias on search fails for your case? I think we need to think of a better solution. 

BTW, what version of the package are you using?

openvast:
laravel 5.0, datatable 1.107

yajra:
BTW, I am receiving your emails it might have took longer to load due too images. Will try to fix this on some other way as this will break some of the installations. Let's keep this open until we have a better solution. Thanks!

Also, I suggest you use `filterColumn` atm to fix the where clause issue?

openvast:
are you try my change?if have any errors, tell me ,let us solve that .if don't have any bug ,why not to use this fix.
if the fix effect select like "DB::raw('(SELECT table2.id FROM table as t ...) as file_id')", than you can rewrite the function extractColumnName, like : 
public function extractColumnName($str)
    {
        dd($str);
        $matches = explode(' as ', Str::lower($str));

        if ( ! empty($matches)) {
            if (Str::contains($str, 'DB::raw')) {
                return array_pop($matches);
            } else {
                return array_shift($matches);
            }
        } elseif (strpos($str, '.')) {
            $array = explode('.', $str);

            return array_pop($array);
        }

        return $str;
    }

yajra:
DB::raw will not be passed on the string so that will probably not work.

yajra:
BTW, I think this PR will work when searching but will still fail when ordering is called? Have you tried ordering your table using this patch?

openvast:
the problem :  how to decision a column name to search. OK.i try 。。。。
i tried ordering my table using this patch (1 minute ago), don't have bug。。。

yajra:
I think I found a better way of doing this. When searching, `array_shift` is indeed needed but when sorting, `array_pop` is needed since alias works on ordering... Lets try to make a patch based on this theory but need to be careful since I see this as a breaking change. http://stackoverflow.com/questions/942571/using-column-alias-in-where-clause-of-mysql-query-produces-an-error

openvast:
this is a good new, i can stop the try and do my program
wait for you new version,^_^

openvast:
my ideas .
$column = $this->setupColumnName($i);


$i     is int index ,i suggest  use column name do the key , like

$column = $this->setupColumnName('columnName');


function setupColumnName($key) {
     $columns = [
            'columnName1'  => 'name';
            'columnName2'  => 'sex';
            'columnName3'  => 'age';
      ];
      return $columns[$key];
}

PR #275, opened 2015-12-02
use query->leftJoin in datatables

openvast:
line 107.


yajra:
Thanks released on v5.12.2!

PR #276, opened 2015-12-02
use select one column, one column one search 

openvast:
but another bug is :i search quickly，the datatables will let my user auth logout, what happen?

### DIFF
--- a/src/yajra/Datatables/Request.php
+++ b/src/yajra/Datatables/Request.php
@@ -134,7 +134,8 @@ class Request extends IlluminateRequest
     {
         $columns = $this->get('columns');
         if ($column_search) {
-            return $columns[$i]['searchable'] == 'true' && $columns[$i]['search']['value'] != '' && ! empty($columns[$i]['name']);
+            return $columns[$i]['searchable'] == 'true' && $columns[$i]['search']['value'] != '';
+            //return $columns[$i]['searchable'] == 'true' && $columns[$i]['search']['value'] != '' && ! empty($columns[$i]['name']);
         }
 
         return $columns[$i]['searchable'] == 'true';


yajra:
You mean you got logged out? This issue was mostly raised when using `artisan serve`. If you're using serve then I suggest you change your environment to some LAMPP/WAMP stack.

src/yajra/Datatables/Request.php
```diff
@@ -134,7 +134,8 @@ public function isColumnSearchable($i, $column_search = true)
     {
         $columns = $this->get('columns');
         if ($column_search) {
-            return $columns[$i]['searchable'] == 'true' && $columns[$i]['search']['value'] != '' && ! empty($columns[$i]['name']);
+            return $columns[$i]['searchable'] == 'true' && $columns[$i]['search']['value'] != '';
```
yajra: Can you please elaborate the need to remove the check for name index? We need the name index for searching and ordering of queries so it's important to check it before hand. Thanks!

openvast:
I use server is xammp ,
remove the check for name index : the "function isColumnSearchable" is check a column is searchable or not, remove that not effect the "searching and ordering", i test this change in my program, do not  have any problems .

datatables::of($query)->make();
nullname

datatables::of($query)->make(true);
notnullname

yajra:
I see.. since we have a `setupColumnName` method that magically fetch the proper column name then we can indeed remove the check for the name index.

```php
            if ($this->request->isColumnSearchable($i)) {
                $column  = $this->setupColumnName($i);
```

openvast:
haha .you found it

yajra:
Yeah makes sense now. ^_^ Thanks! Released on v5.12.3

PR #279, opened 2015-12-02
get the columnName

openvast:
see the post/GET param

-------------------------------------------
columns[0][data]	
type_name
columns[0][name]	
columns[0][orderable]	
true
columns[0][search][regex]	
false
columns[0][search][value]	
columns[0][searchable]	
true
columns[1][data]	
realname
columns[1][name]	
columns[1][orderable]	
true
columns[1][search][regex]	
false
columns[1][search][value]	
ad
columns[1][searchable]	
true
columns[2][data]	
add_time
columns[2][name]	
columns[2][orderable]	
true
columns[2][search][regex]	
false
columns[2][search][value]	
columns[2][searchable]	
false

-------------------------------------------

data and name ....

### DIFF
--- a/src/yajra/Datatables/Request.php
+++ b/src/yajra/Datatables/Request.php
@@ -158,7 +158,7 @@ class Request extends IlluminateRequest
      */
     public function columnName($i)
     {
-        return $this->get('columns')[$i]['name'];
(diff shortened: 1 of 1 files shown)


openvast:
this patch have bug . use this in (datatable::of(query)->make(true);) is right, 
in (datatable::of(query)->make();) has bug , datatable can't draw, how to fix this bug？i think you has bester way

Issue #212, opened 2015-10-13
HTML Builder with custom javascript template and Html template


minkbear:
First, Thanks for your effort for this package. It's saved my time much.
It's the best if there is blade template to be customize.
For example: https://github.com/Chumper/Datatable/tree/master/src/views

yajra:
@minkbear, I think this is good addition to the package :+1: . Will try to add this on future releases. Currently occupied atm so it may take some time to be implemented. Thanks!

Btw, can you explain how the js template was used? Any proposed workflow how to add this? Haven't used Chumper yet.

yajra:
Upcoming v6.0 would now allow the script to load from a template. Kindly check it out if you time. Thanks!

minkbear:
@yajra thanks

PR #298, opened 2015-12-10
Separate builder script template to config

vladkucherov:


vladkucherov:
@yajra The test doesn't pass because phpunit can't find the config for it, and in return, gets null. can you see why?

yajra:
@vladkucherov, great PR :100: ! You need to mock the Config class to make the test pass. Or maybe just use `config()` function helper for easier mocking? Maybe something like below:

```php
// Builder.php
return sprintf(config('datatables.script_template'), $this->tableAttributes['id'], $parameters);

// tests/helper.php
function config($path) {
    switch($path) {
        case 'datatables.script_template':
            return '(function(window,$){window.LaravelDataTables=window.LaravelDataTables||{};window.LaravelDataTables["%1$s"]=$("#%1$s").DataTable(%2$s);})(window,jQuery);';
    }
}
```

On the other hand, I think we can also convert your PR and load a view instead? This way, we can also address issue #212. Something like:

```php
// Builder.php
return sprintf(
  $this->view->make(config('datatables.script_template', 'datatables::script'))->render(),
  $this->tableAttributes['id'], $parameters
);

// script.blade.php
(function(window,$){window.LaravelDataTables=window.LaravelDataTables||{};window.LaravelDataTables["%1$s"]=$("#%1$s").DataTable(%2$s);})(window,jQuery);

// config/datatables.php
'script_template' => 'datatables::script'
```

yajra:
@vladkucherov, sorry for taking over your PR. I went ahead and implemented the solution I proposed where we use a blade template to be used for script. Credits to your idea on how implement custom template for js. 

Btw, if it's ok, can you please test the latest commit on develop/6.0 branch if it works for you? Thanks a lot!

yajra:
Also, I've added a `setTemplate('path.to.view')` method, where we can override the default view to be used for js template. FYI.

```php
public function html()
    {
        return $this->builder()
            ->columns([
                'id',
                'name',
                'email',
                'created_at',
                'updated_at',
            ])
            ->parameters([
                'dom' => 'Bfrtip',
                'buttons' => ['csv', 'excel', 'pdf', 'print', 'reset', 'reload'],
            ])
            ->setTemplate('custom.script');
    }
```

vladkucherov:
I tested it out, and looks fine now ! :+1:

PR #300, opened 2015-12-13
Render column parameter with callback

vladkucherov:


yajra:
@vladkucherov, thanks for this PR. Can you please provide some sample snippet on how this works? Will try to test this as soon as I can. Thanks!

vladkucherov:
This goes in the DataTables Service
```php
return $this->builder()
    ->columns([
        'id',
        'name',
        'description',
        'created_at'
    ])
    ->addAction([
        'render' => null // this is by default. if null, will use the default view from config
    ]);
```

Another use case:
```php
return $this->builder()
    ->columns([
        'id',
        'name',
        'description',
        'created_at'
    ])
    ->addAction([
        'render' => function () {
            return 'the result you wish to return instead of a view';
        }
    ]);
```

src/Html/Column.php
```diff
@@ -20,6 +20,36 @@ public function __construct($attributes = [])
...
+     * @return string
+     */
+    public function parseRender($value)
+    {
+        $value = $value ?: $this->config->get('datatables.render_template', 'datatables::action');
+
+        if(is_callable($value)) {
+            $value = value($value);
+        } else {
+            $value = view($value)->render();
```
yajra: Maybe we should use `$this->view->make($value)->render()` since we already injected the view factory here.

src/Html/Column.php
```diff
@@ -20,6 +20,36 @@ public function __construct($attributes = [])
...
         $attributes['searchable'] = isset($attributes['searchable']) ? $attributes['searchable'] : true;
 
+        // Allow methods override attribute value
+        foreach($attributes as $attribute => $value) {
+            $method = 'parse' . ucfirst(strtolower($attribute));
+            if(method_exists($this, $method)) {
+                $attributes[$attribute] = $this->$method($value);
+            }
+        }
+
```
yajra: Can you please provide some snippets on how to implement/use this?

vladkucherov:
Hi @yajra 
Regarding the config & view. It's not being injected in the Column class. It's being injected in the Builder. Any ideas?

Regarding the `parseColumn` it's being used in the Column class, not Builder. you think it should be move to the Builder?

yajra:
@vladkucherov, yeah my bad. View is not on Column class. Will test this today and let you know what I may find. Thanks!

src/Html/Column.php
```diff
@@ -20,6 +20,36 @@ public function __construct($attributes = [])
...
+     * Parse Render
+     *
+     * @param $value
+     *
+     * @return string
+     */
+    public function parseRender($value)
+    {
+        $value = $value ?: $this->config->get('datatables.render_template', 'datatables::action');
+
```
yajra: `$this->config->get()`.
Tests will fail because we don't have config here. Maybe we can use `app('view')` and `app('config')` and just mock the app function on tests.

yajra:
Suggestions on Column class. This will allow the following scenario:
```php
'field1' => ['render' => 'any string'],
'field2' => ['render' => 'path.to.view'],
'field3' => ['render' => function() {
	return 'function(data,type,full,meta){ return data; }';
}],
```

```php
/**
 * Parse render attribute.
 *
 * @param \Closure|string $value
 * @return string|null
 */
public function parseRender($value)
{
    $view = app('view');

    if (is_callable($value)) {
        return value($value);
    } elseif ($view->exists($value)) {
        return $view->make($value)->render();
    }

    return $value ? $this->parseRenderAsString($value) : null;
}

/**
 * Display render value as is.
 *
 * @param string $value
 * @return string
 */
private function parseRenderAsString($value)
{
    return "function(data,type,full,meta){return '$value';}";
}
```

PR #309, opened 2015-12-17
get order column name from the request

WilliamPeralta:
so we can take the field in the database in any order, because in this way the name of the field does not depend on the order of the query sql but the parameter's request "column"


yajra:
@WilliamPeralta, thanks for this PR. Released on v5.12.5.

PR #358, opened 2016-01-27
Change how regex code is generated after a column search.

ansient:
Before this change, regex statements were generated with the MySQL operator 'REGEXP', which were leading to a Oracle error.
Now it uses the Oracle Database SQL function 'REGEXP_LIKE', as written in the Oracle Database Online Documentation, 10g Release 2 (10.2) .

### DIFF
--- a/src/Engines/BaseEngine.php
+++ b/src/Engines/BaseEngine.php
@@ -855,4 +855,14 @@ abstract class BaseEngine implements DataTableEngineContract
 
         return $this;
     }
+	
+	/**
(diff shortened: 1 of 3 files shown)


yajra:
@ansient, thanks for this PR. The original PR #152 for regex support is indeed only for MySQL and SQLite. Merging this will hence break those installations. Maybe we should add some check of what DB the framework is using then use the appropriate regex command to use?

ansient:
Oh, sorry, I was too hasty.  
Yeah, a check would work well, let me try something.

PR #372, opened 2016-02-03
Update QueryBuilderEngine.php

Worm:
public function count(). If I use in SELECT same custom values, for example (count(samevalue)/count(other_same_value) as `3value`) and use in "group by" or  "order by" this value, I have error: Column not found: 1054 Unknown column '3value' in 'order clause' :(

### DIFF
--- a/src/Engines/QueryBuilderEngine.php
+++ b/src/Engines/QueryBuilderEngine.php
@@ -85,7 +85,7 @@ class QueryBuilderEngine extends BaseEngine implements DataTableEngineContract
         $myQuery = clone $this->query;
         // if its a normal query ( no union, having and distinct word )
         // replace the select with static text to improve performance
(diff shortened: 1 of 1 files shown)


yajra:
Thanks for the PR but do we really need to add `as` in the exceptions? I think `order by` and `group by` is enough? Will your query work when we remove the `as` in exceptions?

Worm:
Yes, my query will be work, but I'm not sure that you want to remove "as". If we use custom select "var" in "where" we see error. mmm?

yajra:
As far as I know, using the alias column via `as` for ordering works well for most database. What makes your query different is that you use a `group by` and `order by` clause which probably causes count to fail?

And maybe most of the users are using alias on their queries. Adding it in the exceptions may cause some performance issues for them?

Worm:
Ok, delete "as", confirmed:)

yajra:
Thanks! released on v6.3.2 :+1:

PR #381, opened 2016-02-09
Global search on eager loaded models

ikerasLT:
The column name should be stated as
`name: relation.column_name`

### DIFF
--- a/src/Engines/QueryBuilderEngine.php
+++ b/src/Engines/QueryBuilderEngine.php
@@ -99,8 +99,10 @@ class QueryBuilderEngine extends BaseEngine implements DataTableEngineContract
      */
     public function filtering()
     {
+        $eagerLoads = array_keys($this->query->getEagerLoads());
+
         $this->query->where(
-            function ($query) {
+            function ($query) use ($eagerLoads) {
                 $keyword = $this->setupKeyword($this->request->keyword());
                 foreach ($this->request->searchableColumnIndex() as $index) {
                     $columnName = $this->setupColumnName($index);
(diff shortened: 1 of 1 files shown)


yajra:
@ikerasLT, thanks for the PR will review and test this as soon as I can. This will definitely help a lot and is a desired feature of the package! :beers:

src/Engines/QueryBuilderEngine.php
```diff
@@ -99,8 +99,10 @@ public function count()
      */
     public function filtering()
     {
+        $eagerLoads = array_keys($this->query->getEagerLoads());
```
yajra: This only works Model but will fail on Query Builder. Maybe we should override filtering on EloquentEngine instead of QueryBuilderEngine.

src/Engines/QueryBuilderEngine.php
```diff
@@ -186,6 +200,24 @@ protected function compileGlobalSearch($query, $column, $keyword)
     }
 
     /**
+     * Add relation query on global search
+     *
+     * @param mixed $query
+     * @param string $column
+     * @param string $keyword
+     */
+    protected function CompileRelationSearch($query, $relation, $column, $keyword)
```
yajra: Please use `compileRelationSearch` instead of `CompileRelationSearch`.

src/Engines/QueryBuilderEngine.php
```diff
@@ -186,6 +200,24 @@ protected function compileGlobalSearch($query, $column, $keyword)
     }
 
     /**
+     * Add relation query on global search
```
yajra: Missing period. ^_^

src/Engines/QueryBuilderEngine.php
yajra: Or maybe we can extract this to own method that handles the exception when using query builder.

```php
    $eagerLoads = $this->getEagerLoads();
----------
    /**
     * Get eager loads key if eloquent.
     *
     * @return array
     */
    private function getEagerLoads()
    {
        if ($this->query_type == 'eloquent') {
            return array_keys($this->query->getEagerLoads());
        }

        return [];
    }
```

yajra:
@ikerasLT thanks for this initiative. I went ahead and implemented some fixes to your PR. Released on v6.4.0. Thanks again! :beers:

PR #384, opened 2016-02-10
Search and order by raw columns

rickatultraware:
Make it posible to seacrh and order with raw column name.
You have to provide raw_ prefix in the name of the column
```
columns: [
  {data: 'concat', name: 'raw_concat(col1,col2)'},
]
```

### DIFF
--- a/src/Engines/QueryBuilderEngine.php
+++ b/src/Engines/QueryBuilderEngine.php
@@ -243,6 +243,9 @@ class QueryBuilderEngine extends BaseEngine implements DataTableEngineContract
      */
     public function castColumn($column)
     {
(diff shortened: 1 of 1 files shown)


yajra:
Need to review this further. Do we really need to allow raw statements? Not sure, but it may cause us some sql injections attacks since it's a raw statement?

On the other hand, I think you can also achieve the same result using `filterColumn` and `orderColumn` method?

rickatultraware:
You are right. it may cause us some sql injections attacks. 
Because i made a class that reconstruct the request this wasn't really a problem for me. 
I archived it bij using the ```filterColumn``` method.

If you are interested in the class that reconstruct the request just let me know.

PR #392, opened 2016-02-11
Fix filtering in nested columns of Collections

FabryB:
With this small change filtering is working even with nested columns, using dot notation.


yajra:
Thanks a lot! Released on v6.4.4. :+1:

PR #399, opened 2016-02-17
Allow edit columns for nested arrays

ramilexe:
For example you have `User` and `Phone` models. `User` has one `Phone`.  `Phone` has `name` field
And you want to edit phone.name field. My pull request allow to do it:

```
$users = User::with('phone');
\Datatables::of($users)
  ->editColumn('phone.name', function($user) {
    return 'Name of phone is: ' . $user->phone->name;
  })
  ->make(true);
```

### DIFF
--- a/src/Processors/DataProcessor.php
+++ b/src/Processors/DataProcessor.php
@@ -2,6 +2,7 @@
 
 namespace Yajra\Datatables\Processors;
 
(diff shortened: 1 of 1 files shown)


yajra:
Thanks a lot @ramilexe! Released on v6.4.5

PR #400, opened 2016-02-17
In the refactoring process, of the column search system, an error rel…

ansient:
…ated to a case-sensitivity check led to a wrong interpretation and therefore to an incorrect behaviour of the logic.

If the isCaseInsensitive function is true, the compileColumnSearch function is called with the flag caseSensitive set to true, clearly wrong.
The same issue occurs if the isCaseInsensitive function is false.

Now the boolean flag caseSensitive is set correctly for both if-else blocks.


yajra:
Thanks a lot @ansient! Released on v6.4.5

Issue #217, opened 2015-10-20
filterColumn 


luciantugui:
can I filter data by exact match? it seems like filterColumn adds wildcard % all the time. Is it possible to disable wildcard when calling filterColumn?

this also adds the wildcard:
->filterColumn('price_gross', 'whereRaw', 'price_gross' = ?', ["$1"])

yajra:
`filterColumn` uses wildcard search settings. If you want to do an exact match, then you would have to do it manually by using request object and get the search value. Using `$1` variable will yield to `%search%`. Try something like below:
```php
->filterColumn('price_gross', 'whereRaw', 'price_gross' = ?', [$request->get('search')['value']])
```

SuperlativeEntity:
```
            $data = datatables()->of($members);

            $data->filterColumn('member_number', function ($query, $keyword)
            {
                $query->where('member_number', '=', $keyword);
            });

            return $data->toJson();
```

Issue #191, opened 2015-09-17
Search with boolean doesn't work well


joanebrown:
Hi,

I'm trying to search the first column of my dataTable (the active column), and I use this:

    table.DataTable().columns(0).search("0");

But instead of return all the rows with the active column = 0, this return me all rows...

It's weird, but if I use:

    table.DataTable().columns(0).search("1");

Works well, and I only see the active rows...

Maybe is a bug?

yajra:
Are you using regex? Kindly check the queries generated on dev tools and verify if it's correct. There used to be a bug when searching zero value but this was fixed sometime ago as far as I remember.

joanebrown:
No, I onle use this code... In the Dev Tools I see all well :S 

When I use:

    table.DataTable().columns(0).search("%0%");

All works well, Why?


**BTW**: How can I implement a search if my column is produced by editColumn? (my "product_id" but I edit this with the serial number of the product: product->serialNumber). So when I search... I don't found nothing, because datatables search by the ID... Sorry for my english T_T

yajra:
I tried column search with zero value and it works well. Try executing the generated sql manually and check if it does work. Demo for column search http://datatables.yajrabox.com/eloquent/multi-filter-select seems to work well too.

For editColumn, searching on it's edited value may not work since we are using a database query and it's not the value stored on it. For your case, you should use join statements (I think you are using eager loading?). Searching on eager loaded models is not supported atm.

joanebrown:
Hi Yajra,

I tried with the SQL directly, and works well...

    select * from products where active = '1';

But doesn't work with:

    table.DataTable().columns(0).search("1");
And "works" with:

    table.DataTable().columns(0).search("%0%");

I say "works" inside quotes because, if the column have the "01" value, i get this too, with the "0". (because % is a wildcard character...)

Why is happen that?

BTW, the next sentence:

    $.fn.dataTable.util.escapeRegex($(this).val());

return me the next error:

    Uncaught TypeError: Cannot read property 'escapeRegex' of undefined

yajra:
Is that the sql generated on ajax request? Can you dump the ajax output please?

joanebrown:
I use AJAX on my datatables, yes. Of course!
In the next ajax output, i try to search the element with id 7, but i get the id 17 too:

[...]

PR #440, opened 2016-03-08
Added Closure support for filterColumn method

codewizz:
Hello,

with the following PR I have added a closure support for filterColumn method as a more advanced and flexible way of customising things.

Instead of writing like this:
```php
->filterColumn('id', 'whereRaw', 'interface_strings.interface_string_id = ?', [$request->get('columns')[0]['search']['value']])
```
we can now use like this ($keyword supplied to a closure is without wildcards - especially handy when we want to do an exact match):
```php
->filterColumn('id', function($query, $keyword) {
    $query->where('interface_strings.interface_string_id', $keyword);
})
```

Exact match issues specified in #217, #191 should be covered by this PR.

Looking forward!

### DIFF
--- a/src/Engines/QueryBuilderEngine.php
+++ b/src/Engines/QueryBuilderEngine.php
@@ -102,20 +102,33 @@ class QueryBuilderEngine extends BaseEngine
     {
         $this->query->where(
             function ($query) {
-                $keyword = $this->setupKeyword($this->request->keyword());
+                $globalKeyword = $this->setupKeyword($this->request->keyword());
+                $queryBuilder = $this->getQueryBuilder($query);
+
                 foreach ($this->request->searchableColumnIndex() as $index) {
                     $columnName = $this->getColumnName($index);
-
+                    // check if custom column filtering is applied
                     if (isset($this->columnDef['filter'][$columnName])) {
-                        $method     = Helper::getOrMethod($this->columnDef['filter'][$columnName]['method']);
-                        $parameters = $this->columnDef['filter'][$columnName]['parameters'];
-                        $this->compileColumnQuery(
-                            $this->getQueryBuilder($query),
-                            $method,
-                            $parameters,
-                            $columnName,
-                            $keyword
-                        );
+                        $columnDef = $this->columnDef['filter'][$columnName];
+                        // check if global search should be applied for the specific column
+                        $applyGlobalSearch = count($columnDef['parameters']) == 0 || end($columnDef['parameters']) !== false;
+                        if (!$applyGlobalSearch) {
+                            continue;
+                        }
+
+                        if ($columnDef['method'] instanceof Closure) {
+                            $whereQuery = $queryBuilder->newQuery();
+                            call_user_func_array($columnDef['method'], [$whereQuery, $this->request->keyword()]);
+                            $queryBuilder->addNestedWhereQuery($whereQuery, 'or');
+                        } else {
+                            $this->compileColumnQuery(
+                                $queryBuilder,
+                                Helper::getOrMethod($columnDef['method']),
+                                $columnDef['parameters'],
+                                $columnName,
+                                $this->request->keyword()
+                            );
+                        }
                     } else {
                         if (count(explode('.', $columnName)) > 1) {
                             $eagerLoads     = $this->getEagerLoads();
@@ -124,16 +137,16 @@ class QueryBuilderEngine extends BaseEngine
                             $relation       = implode('.', $parts);
                             if (in_array($relation, $eagerLoads)) {
                                 $this->compileRelationSearch(
-                                    $this->getQueryBuilder($query),
+                                    $queryBuilder,
                                     $relation,
                                     $relationColumn,
-                                    $keyword
+                                    $globalKeyword
                                 );
                             } else {
-                                $this->compileGlobalSearch($this->getQueryBuilder($query), $columnName, $keyword);
+                                $this->compileGlobalSearch($queryBuilder, $columnName, $globalKeyword);
                             }
                         } else {
-                            $this->compileGlobalSearch($this->getQueryBuilder($query), $columnName, $keyword);
+                            $this->compileGlobalSearch($queryBuilder, $columnName, $globalKeyword);
                         }
                     }
 
@@ -147,7 +160,7 @@ class QueryBuilderEngine extends BaseEngine
      * Perform filter column on selected field.
      *
      * @param mixed $query
-     * @param string $method
+     * @param string|Closure $method
      * @param mixed $parameters
      * @param string $column
      * @param string $keyword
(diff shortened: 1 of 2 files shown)


yajra:
@codewizz, great PR! :100:  Will test this as soon as I can. Thanks a lot!

yajra:
@codewizz, I just reviewed your PR and implementing this may cause some issue for global search. Some Issues found atm:

**Issue 1.** Global search also use `filterColumn` method so you need to implement it there too. I already done this part. See snippet below:
```php
    /**
     * Perform global search.
     *
     * @return void
     */
    public function filtering()
    {
        $this->query->where(
            function ($query) {
                $keyword = $this->setupKeyword($this->request->keyword());
                foreach ($this->request->searchableColumnIndex() as $i) {
                    $column = $this->getColumnName($i);

                    if (isset($this->columnDef['filter'][$column])) {
                        $method = $this->columnDef['filter'][$column]['method'];
                        if (! ($method instanceof Closure)) {
                            $method     = Helper::getOrMethod($method);
                            $parameters = $this->columnDef['filter'][$column]['parameters'];
                            $keyword    = $this->getSearchKeyword($i);
                            $this->compileColumnQuery(
                                $this->getQueryBuilder(),
                                $method,
                                $parameters,
                                $column,
                                $keyword
                            );
                        } else {
                            $keyword = $this->getSearchKeyword($i, true);
                            call_user_func_array($method, [$this->getQueryBuilder(), $keyword]);
                        }
                    } else {
                        if (count(explode('.', $column)) > 1) {
                            $eagerLoads     = $this->getEagerLoads();
                            $parts          = explode('.', $column);
                            $relationColumn = array_pop($parts);
                            $relation       = implode('.', $parts);
                            if (in_array($relation, $eagerLoads)) {
                                $this->compileRelationSearch(
                                    $this->getQueryBuilder($query),
                                    $relation,
                                    $relationColumn,
                                    $keyword
                                );
                            } else {
                                $this->compileGlobalSearch($this->getQueryBuilder($query), $column, $keyword);
                            }
                        } else {
                            $this->compileGlobalSearch($this->getQueryBuilder($query), $column, $keyword);
                        }
                    }

                    $this->isFilterApplied = true;
                }
            }
        );
    }
````

**Issue 2.** When using global search, we need to use the `or` counter part method of the query. For instance, `where('id', 1)` should become `orWhere('id', 1)`. On the contrary, column search can be either **and/or** query (but it's more correct to use **and** query). Like on [filterColumn demo](http://datatables.yajrabox.com/eloquent/post-column-search) which is what I used for testing, we can specify it an `orWhereRaw` method to work with global search. However, column search may not work properly with that.
```php
// Works with global search but not on column search
        return Datatables::of($users)
            ->filterColumn('user_id', function($query, $keyword) {
                $query->orWhereRaw("CONCAT(users.id,'-',users.id) like ?", ["%{$keyword}%"]);
            })
            ->make(true);

// Works with column search but not on global search
        return Datatables::of($users)
            ->filterColumn('user_id', function($query, $keyword) {
                $query->whereRaw("CONCAT(users.id,'-',users.id) like ?", ["%{$keyword}%"]);
            })
            ->make(true);
```

codewizz:
@yajra oh right, I did not consider this for a global search, as I was working on column filtering when searched for the following funcionality, everything you wrote does make sense indeed. I will do some testing on the orWhere thing, I guess we can easly contain / group orWhere part within a library.

Also, just a thought on the following use-case scenario:
```php
->filterColumn('id', 'whereRaw', 'interface_strings.interface_string_id = ?', ['$1'])
```

if you think to keep this implementation in the future (even I do prefer the closure one - more laravelish style), maybe it's worth to return raw keyword as $1 variable here, like I am doing now in a closure implementation. Usually people will want to do an exact match if writing in this form, and if LIKE search is needed we can easly add ['%$i%'] .... I guess it's worth considering.

I will report back shortly after doing some tests.

codewizz:
@yajra a quick question, should I include your snippet into my PR, or you will add this separately?

yajra:
I think you can add it on your PR. not sure how to insert it here. Thanks!

codewizz:
@yajra any thoughts on giving back a raw keyword for ['$1'] ?

yajra:
I think the your approach of raw keyword is fine since we can easily add a wildcard search if we wanted to. The only thing stopping as here is how can we convert the query from closure to an equivalent `or` method.

yajra:
> if you think to keep this implementation in the future (even I do prefer the closure one - more laravelish style), maybe it's worth to return raw keyword as $1 variable here, like I am doing now in a closure implementation. Usually people will want to do an exact match if writing in this form, and if LIKE search is needed we can easly add ['%$i%'] .... I guess it's worth considering.

I guess it's worth considering to return raw on `$1` but it's a breaking change so we need to be careful on it.

codewizz:
@yajra is there any use-case where we would want to apply a global filter on top of an overriden column filter?

So if we have a table with the following columns
```
id
title
message
```

```php
return Datatables::of($builder)
      ->filterColumn('id', function($query, $keyword) {
             $query->where('id', $keyword);
       }) ->make(true);
```

Does it make sense to apply a global search for ``id`` column here if it's being overridden anyway?

I would expect here my SQL query ``where`` part would look something like:

```sql
where (`title` LIKE '%keyword%' or `message` LIKE '%keyword%') and (`id` = '4')
```
but generated MySQL query looks like:

```sql
where ((`id` = 4) or `title` LIKE '%keyword%' or `message` LIKE '%keyword%') and (`id` = '4')
```

So the question is do we want this? Maybe we should apply a global search only for fields that are not overridden?

yajra:
The most common use case of `filterColumn` on global search is like on demo where columns were concatenated. For instance, you may want to search on users `full_name` from concatenating `first_name` and `last_name` with space. Using normal global search won't be able to do that.

codewizz:
@yajra are you refering to this demo: http://datatables.yajrabox.com/fluent/advance-filter ?

yajra:
Yes, and something like this demo too http://datatables.yajrabox.com/eloquent/post-column-search.

Actually we can make closure work if column search and global search is not use at the same time. The issue occurs only when both search are used.

codewizz:
@yajra is there a way to see SQL query generated (when both column and global filters are used) for http://datatables.yajrabox.com/eloquent/post-column-search demo? I guess the debug mode is turned off at the moment, as I could not see query returned in XHR response.

yajra:
@codewizz, I just enabled debug mode on demo. You should be able to see the queries now.

codewizz:
@yajra thanks, I kinda see your point now.

```sql
select CONCAT(users.id,'-',users.id) as user_id, `name`, `email`, `created_at`, `updated_at` from `users` where (CONCAT(users.id,'-',users.id) like ? or LOWER(`name`) LIKE ? or LOWER(`email`) LIKE ? or LOWER(`created_at`) LIKE ? or LOWER(`updated_at`) LIKE ?) and CONCAT(users.id,'-',users.id) like ? order by `user_id` asc limit 10 offset 0
```

with the following bindings: ``["%Roel%","%roel%","%roel%","%roel%","%roel%","%10%"]``

p.s. a bit un-related but why are all other bindings in a lower case except the first one, maybe it's a bug (original input was "Roel")? This would fail when doing search using a case sensitive ``utf8_bin`` collation.

Is there any funcionality now to disable a global search for one or another column?

If no maybe I could add a third (optional) parameter to specify this (in case we don't want a global filter be applied for the overriden filter column - small optimisation + a bit of flexibility)?

```php
      ->filterColumn('id', function($query, $keyword) {
             $query->where('id', $keyword);
       }, false)  // <--- default would be true ($applyGlobalSearch = true)
```

also I will fix the default behaviour to be applied for closured filter columns.

Looking forward!

yajra:
I think your proposal for a third bool parameter is good. I agree to proceed with that. 

For the bindings, you are correct that there seems to be a bug there. `$1` keyword should return a lower case value too. Will review it too. Thanks!

codewizz:
> For the bindings, you are correct that there seems to be a bug there. $1 keyword should return a lower case value too. Will review it too. Thanks!

I would suggest leaving an original input keyword value for all bindings, otherwise user will not be able do a binary search and match text with an upper case.

codewizz:
@yajra I have just update my PR with the things we did discuss here. Looking forward!

yajra:
Thanks @codewizz, will review this as soon as I can.

src/Engines/QueryBuilderEngine.php
```diff
@@ -102,20 +102,33 @@ public function filtering()
...
+                        // check if global search should be applied for the specific column
+                        $applyGlobalSearch = count($columnDef['parameters']) == 0 || end($columnDef['parameters']) !== false;
+                        if (!$applyGlobalSearch) {
+                            continue;
+                        }
+
+                        if ($columnDef['method'] instanceof Closure) {
+                            $whereQuery = $queryBuilder->newQuery();
+                            call_user_func_array($columnDef['method'], [$whereQuery, $this->request->keyword()]);
+                            $queryBuilder->addNestedWhereQuery($whereQuery);
```
yajra: Great implementation you've done here. We just need to add `or` as 2nd param and our problem is solved.
```php
$queryBuilder->addNestedWhereQuery($whereQuery, 'or');
```

src/Engines/QueryBuilderEngine.php
```diff
@@ -102,20 +102,33 @@ public function filtering()
...
-                        $this->compileColumnQuery(
-                            $this->getQueryBuilder($query),
-                            $method,
-                            $parameters,
-                            $columnName,
-                            $keyword
-                        );
+                        $columnDef = $this->columnDef['filter'][$columnName];
+                        // check if global search should be applied for the specific column
+                        $applyGlobalSearch = count($columnDef['parameters']) == 0 || end($columnDef['parameters']) !== false;
```
yajra: No need to filter out global search when we use `$queryBuilder->addNestedWhereQuery($whereQuery, 'or');`. This condition can be removed now.

codewizz: Yes, but $applyGlobalSearch was done intentionally, so I could skip LIKE part for the particular column, when defining a custom filter. Most of the time you might not want to apply global filter for a custom filter column, even it would work with an added "or" but there might be cases where it becomes difficult to do an exact match. I would recommend to leave this option as is for the user to decide and be more flexible on defining the flow of custom filtering.

usage example code, when the applyGlobalSearch is not applied (this does guarantee an exact match for the column):
```php
      ->filterColumn('id', function($query, $keyword) {
             $query->where('id', $keyword);
       }, false)  // override a global search (disable) for the custom filter 
```

Cheers!

src/Engines/QueryBuilderEngine.php
codewizz: Indeed. The best thing is we will be able to group multiple where's in the nested where group (and do some more advanced things in where part) easly:
```php
       ->filterColumn('id', function($query, $keyword) {
             $query->where('user_id', $keyword)->where('type', 'customer');
       }, false)  // override a global search (disable) for the custom filter 
```

p.s. just a thought, maybe it would make sense to give a globalKeyword as a third Closure parameter?

codewizz:
@yajra did add a missing ``or`` for ``addNestedWhereQuery``. Thanks for keeping an eagle eye!

yajra:
@codewizz, thanks a lot for this PR. Released on v6.8.0. :beers:

codewizz:
@yajra it was a pleasure :beers: . Looking forward to contribute more in the future!

eelco2k:
I found this thread regarding the implementation of the Closure filterColumn method, i'm experiencing some issues with a query select "as alias" and filtering on that column name.

i already opened this as a separe issue here: [https://github.com/yajra/laravel-datatables/issues/572](https://github.com/yajra/laravel-datatables/issues/572)

PR #471, opened 2016-03-18
Add option to create a table footer from builder defined via column def

yajra:
This PR will add an option to add a table footer via Builder. Footer contents will be defined via column definition.

## Usage
```php
$builder->columns([
    'id' => ['footer' => 'footer content']
]);

$dataTable->table([], true); // passing true as second parameter will draw the footer.
```


Trikke:
any plans to include this on a next release?

yajra:
Yes, I just made some updates earlier and need to test on actual code first.. But this will surely be included on the next release along with the other merged PR. 👍

PR #473, opened 2016-03-18
use orHavingRaw vs orWhereRaw

lisa-fehr:
In the internal datatables search/filter: ``orWhereRaw`` wasn't able to find the 'select as' alias for ``$column``. ``orHavingRaw`` appears to have access to the alias.

The code is the same except ``where`` was removed and ``orWhereRaw`` replaced with ``orHavingRaw``

### DIFF
--- a/src/yajra/Datatables/Datatables.php
+++ b/src/yajra/Datatables/Datatables.php
@@ -395,43 +395,40 @@ class Datatables
         }
 
         if (! empty($this->input['search']['value'])) {
-            $this->query->where(function ($query) use ($columns, $input) {
-                for ($i = 0, $c = count($columns); $i < $c; $i++) {
-                    if ($columns[$i]['searchable'] == "true" and ! empty($columns[$i]['name'])) {
-                        $column = $columns[$i]['name'];
+            for ($i = 0, $c = count($columns); $i < $c; $i++) {
+                if ($columns[$i]['searchable'] == "true" and !empty($columns[$i]['name'])) {
+                    $column = $columns[$i]['name'];
+                    if (stripos($column, ' AS ') !== false) {
+                        $column = $this->getColumnName($column);
+                    }
 
-                        if (stripos($column, ' AS ') !== false) {
-                            $column = $this->getColumnName($column);
-                        }
+                    // there's no need to put the prefix unless the column name is prefixed with the table name.
+                    $column = $this->prefixColumn($column);
 
-                        // there's no need to put the prefix unless the column name is prefixed with the table name.
-                        $column = $this->prefixColumn($column);
+                    $keyword = '%' . $input['search']['value'] . '%';
+                    if ($this->isWildcard()) {
+                        $keyword = $this->wildcardLikeString($input['search']['value']);
+                    }
 
-                        $keyword = '%' . $input['search']['value'] . '%';
-                        if ($this->isWildcard()) {
-                            $keyword = $this->wildcardLikeString($input['search']['value']);
-                        }
-
-                        // Check if the database driver is PostgreSQL
-                        // If it is, cast the current column to TEXT datatype
-                        $cast_begin = null;
-                        $cast_end = null;
-                        if ($this->databaseDriver() === 'pgsql') {
-                            $cast_begin = "CAST(";
-                            $cast_end = " as TEXT)";
-                        }
-
-                        // wrap column possibly allow reserved words to be used as column
-                        $column = $this->wrapColumn($column);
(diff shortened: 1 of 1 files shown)


yajra:
@lisa-fehr sorry for the delays. Will try to review/test this as soon as I can. Thanks!

yajra:
@lisa-fehr, I just tested this PR and it seems to work on MySQL but not for Oracle. Not sure on other DB yet as I am only using those 2 DB atm. I think Oracle requires group by clause for having sql to work properly. 

```sql
"Error Code : 979 Error Message : ORA-00979: not a GROUP BY expression Position : 228 Statement : select t2.* from ( select rownum AS "rn", t1.* from (select users.id, users.username, users.email, 'roles', users.confirmed, users.created_at from users having LOWER(users.id) LIKE :p0 or LOWER(users.username) LIKE :p1 or LOWER(users.confirmed) LIKE :p2 order by users.id asc) t1 ) t2 where t2."rn" between 1 and 10 Bindings : [%admis%,%admis%,%admis%,%admis%,%admis%,%admis%] (SQL: select t2.* from ( select rownum AS "rn", t1.* from (select users.id, users.username, users.email, 'roles', users.confirmed, users.created_at from users having LOWER(users.id) LIKE %admis% or LOWER(users.username) LIKE %admis% or LOWER(users.confirmed) LIKE %admis% order by users.id asc) t1 ) t2 where t2."rn" between 1 and 10)"
```

I also just tested SQLITE and it does not work too:
```sql
"SQLSTATE[HY000]: General error: 1 a GROUP BY clause is required before HAVING (SQL: select count(*) as aggregate from (select "users"."id", "users"."username", "users"."email", 'roles', "users"."confirmed", "users"."created_at" from "users" having LOWER(users.id) LIKE %admin% or LOWER(users.username) LIKE %admin% or LOWER(users.confirmed) LIKE %admin%) count_row_table)"
```

lisa-fehr:
@yajra I added a groupBy that doesn't seem to upset mysql. Does it work for Oracle?

I can do a custom override on mine, just thought I'd try to patch it here (if possible).

yajra:
@lisa-fehr, I made a quick test on this and it still fails on Oracle but its working on SQLITE now. Will try to dig more on your implementation when I had the chance.

On the other hand, I think you still need to enclose the `orHavingRaw` inside the query where method. 
```php
$this->query->where(function ($query) use ($columns, $input) {
.....
}
```

This is essential for column search to work properly. Enclosing it in the where query will yield to a query like this when combined with column search: 
```php
where (col1 like '%asd%' or col2 like '%asd%') and col3 like '%asd';
```

Your proposed implementation will yield to something like below which I think may break the correct results?
```php
having col1 like '%asd%' or col2 like '%asd%' and col3 like '%asd';
```

lisa-fehr:
@yajra Keeping it in the ``where`` function ignored the ``orHaving`` for me. 

But that does seem correct, ``orHaving`` isn't contained.
``
group by `name`, `alias` having LOWER(`name`) LIKE ? or LOWER(`alias`) LIKE ?
``

yajra:
@lisa-fehr, I just read this [article](http://www.programmerinterview.com/index.php/database-sql/having-vs-where-clause/) and it seems like it is better to use `where` clause instead of `having` due to some performance issues. Where clause uses index while having does not. It is designed to be used with aggregates which is the reason behind the need for group by.

Quoting from a comment:
>Very good explanation! A little addendum – a follow-up question might be: Is there a disadvantage to using HAVING, especially if there's an equivalent query with WHERE?

>The answer is: There could be performance issues. Queries with the HAVING statement need to filter the result set with O(n) (n being the number of records in the result set) for the HAVING expressions, while WHERE clauses can use indexes and other optimizations, making it possible to have O(log n) or even O(1) execution time complexity.

>The reason for this is exactly the reason why HAVING exists in the first place: The DB generates the result set with the WHERE expression *before* the aggregate function is executed (so it can't be evaluated *after* it). Since there can't be indexes on result sets (they are dynamic, after all), a HAVING filter has to check each result row.

lisa-fehr:
@yajra 
``
Column not found: 1054 Unknown column 'alias' in 'where clause'
``

Seems like the only way I can avoid changing code in ``doFiltering`` is if I make sure the datatable column ``name`` isn't the alias:

``
columns:
0: {title: "Name", data: "name", name: "database.name", orderable: true, class: "name"}
1: {title: "Alias for an edit button", data: "alias", name: "database.id", orderable: false,…}
``

We have some columns that are run through a presenter to create the links. Small edge case. Thanks for checking it over.

PR #482, opened 2016-03-22
Fix: not call statically recover instance of class

juniorb2ss:


juniorb2ss:
For use 

    $request->replace(['start' => null, 'length' => null]); // for disable auto-pagination in collection

    $datatable = new Datatables(
        DatatableRequest::createFromBase($request)
    );

To change request inputs and pass to Datatable instance.

Use datatable with elasticsearch

Issue #423, opened 2016-03-01
Disable smart filtering


brunjick:
Is it possible to disable smart filtering on server side? Because, as I could see, datatables doesn't send any parameter for that, if I set
```
$("#example").DataTable({
    ...
    search: {
        smart: false
    }
    ...
})
```

yajra:
@brunjick, good point. We currently don't have way to disable smart search. Currently we have two config for search but setting both to false will still do smart search and will still perform `like '%search%'`.

```php
    'search'          => [
        'case_insensitive' => false,
        'use_wildcards'    => false,
    ],
``` 

Will add this for enhancement to be configurable via config file. We cannot use the js `smart: false` since this data is not being sent to the server as far as I know. Thanks!

```php
    'search'          => [
        'smart' => true
        'case_insensitive' => false,
        'use_wildcards'    => false,
    ],
```

yajra:
@brunjick, please clarify. What do you mean by smart search? Does it mean **exact search** or you wanted to **disable global search**?

My interpretation is exact search but just wanted to make sure before I try to implement this. Thanks!

brunjick:
Yes I mean exact search. More specifically (https://datatables.net/reference/option/search.smart)[official docs here]

yajra:
Thanks for clarifying. Will try to implement this as soon as I can. :+1:

yajra:
@brunjick, I just created a PR for this feature request. Would you please test if the proposal works for you? Thanks!

brunjick:
Yes, I tested and it worked after fixing one problem. When setting where clause on query object
`$this->query->where($column, 'like', $keyword);`
provided column name is already wrapped in backquote (\`) symbol, so we have to remove them, otherwise query builder will escape them and it becomes wrapped in three backquote symbols. So above code becomes something like this:
```$this->query->where(str_replace('`', '', $column), 'like', $keyword);```
I'm not sure if it's perfect way, but it works now.

Also, is there any chance to implement option to define specific columns for disabling/enabling smart search? for example:
```
 Datatables::of($query)
            ->smart(false, ['column1', 'column2'])
            ->make(true);
```

yajra:
Just wondering why you're column has backtick? Laravel automatically adds it when using query builder so I guess it's not needed?

Will check if it's possible to do it on a column level smart search. Thanks!

brunjick:
No, my column doesn't have one
`$column = $this->connection->getQueryGrammar()->wrap($column);`
[...]

PR #488, opened 2016-03-26
Update CollectionEngine.php to make dropdown-filter work.

amitavmohanty:
Hello Arjay,

Thanks for developing such as a nice plugin. 

I was having an issue when using datatables  with column filters (for dropdown column filter), it was not returning empty record collection.

**\* If ^ and $ present at the beginnning and end of the keyword then do the full match instead of partial match for dorpdown filter. ***

Please check approve the pull request if you think it would improvise the plugin...

Thanks & Regards
Amitav

### DIFF
--- a/src/Engines/CollectionEngine.php
+++ b/src/Engines/CollectionEngine.php
@@ -187,7 +187,12 @@ class CollectionEngine extends BaseEngine
                         $value = Arr::get($data, $column);
 
                         if ($this->isCaseInsensitive()) {
-                            return strpos(Str::lower($value), Str::lower($keyword)) !== false;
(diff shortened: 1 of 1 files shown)


yajra:
@amitavmohanty, thanks for the PR. I just recently released an update that support exact search but it was only implemented on Eloquent/QueryBuilder engine. It would be nice if we can also adapt the approach done on it to ColletionEngine. But note that exact search works on whole query and not on a column level. See this PR for details. https://github.com/yajra/laravel-datatables/pull/452

What I am planning to do on future release is to support column level exact search as discussed on issue #423 which I think is what you're trying to do. If you can, please do the PR as discussed in the issue like:
```php
->smart(false, ['col1', 'col2', ...])
// or maybe something like below since smart was already used to disable smart search.
->exactSearch(['col1', 'col2', ...])
```

PR #552, opened 2016-05-13
Add CAST for Firebird

donnykurnia:
Reference:
http://tracker.firebirdsql.org/browse/CORE-3559

### DIFF
--- a/src/Engines/QueryBuilderEngine.php
+++ b/src/Engines/QueryBuilderEngine.php
@@ -283,6 +283,9 @@ class QueryBuilderEngine extends BaseEngine
         if ($this->database === 'pgsql') {
             $column = 'CAST(' . $column . ' as TEXT)';
         }
+        elseif ($this->database === 'firebird') {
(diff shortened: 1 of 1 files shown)


yajra:
Looks fine for me. One question though, does `VARCHAR(255)` is enough? What if the column has more than 255 bytes? Not using firebird so I'm not really sure about it. Thanks!

donnykurnia:
@yajra In my current project, it's enough. In Firebird, [maximum VARCHAR is 32765] (http://www.firebirdsql.org/manual/migration-mssql-data-types.html)

Firebird don't have TEXT type like in psql, only blob, but blob feature in [laravel driver](https://github.com/jacquestvanzuydam/laravel-firebird) is buggy so I'm avoiding it right now.

yajra:
Ok, will accept this for a start. Thanks!

yajra:
Merged and release on v6.11.2. Thanks! 👍

PR #582, opened 2016-05-30
Add support for laravel-snappy when exporting to PDF.

yajra:
Add support for laravel-snappy when exporting to PDF. 

### How To Use:
- Install [laravel-snappy](https://github.com/barryvdh/laravel-snappy) via `composer require barryvdh/laravel-snappy`
- Register snappy provider: `Barryvdh\Snappy\ServiceProvider::class`.
- Install [wkhtmltopdf](wkhtmltopdf.org/downloads.html). (Requirements of laravel-snappy)
- `use SnappyDataTableTrait` on your DataTable class.
- Done...

Exporting of PDF will now use `laravel-snappy` instead of `laravel-excel` pdf export function which seems to fail most of the time especially when the data is large. Exporting via snappy will be the same as how you print the DataTable.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -25,6 +25,9 @@
         "mockery/mockery": "~0.9",
         "phpunit/phpunit": "~4.0"
     },
+    "suggest": {
+      "barryvdh/laravel-snappy": "Allows exporting dataTable to PDF using the print view."
+    },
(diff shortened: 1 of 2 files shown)


yajra:
Implemented via #626.

Issue #567, opened 2016-05-19
Search not working in HasOne relationship.


mortenlandmark:
Ok I have tried for a while, searched extensively, please help me out.

The table displays correct, but search and sort is only working for rones.RNo, not ords.OrdNo.

So search and sorting is only working for one table. why? Whyyy?

Ord Model:
public function rones()
    {
        return $this->hasOne('App\Rone', 'RNo', 'OrdNo');
    }

Rone Model:
    public function ords()
    {
        return $this->belongsTo('App\Ord', 'OrdNo', 'RNo');
    }

Controller: 
$ords = Ord::with('rones')->get();

Blade:
@extends('layouts.master')
@section('content')
<div class="table-responsive">
    <table width ="100%" class="table table-bordered table-hover table-striped" id="ords-table">
        <thead>
            <tr>
                <th>OrdNo</th>
                <th>RNo</th>
            </tr>
        </thead>
        <tfoot>
            <tr>
                <th>OrdNo</th>
                <th>RNo</th>
            </tr>
        </tfoot>
    </table>
</div>
@stop
@push('scripts')
<script>
$(function() {
    $('#ords-table').DataTable({
        "processing": true,
        "serverSide": true,
        "aoColumnDefs": [
          {
            "mData": null,
            "sDefaultContent": "-",
            "aTargets": [ -1 ]
          }
        ],
        "dom": 'Blfrtip',
        "buttons": [
            'colvis',
        ],
        "ajax": { url: '{!! route('datatablesordsrones.data') !!}' },
        "columns": [
            { data: 'OrdNo', name: 'ords.OrdNo' },
            { data: 'rones.RNo', name: 'rones.RNo', 'defaultContent': " " },
        ],
    });
});
</script>
@endpush

yajra:
Please see this [demo](http://datatables.yajrabox.com/eloquent/joins) and this [one](http://datatables.yajrabox.com/eloquent/relationships) for ref.

mortenlandmark:
Hi.

I have used this demo to set it up:
http://datatables.yajrabox.com/eloquent/relationships

It is set up from two tables, exactly like the example.



But for some reason it's returning the error:
"DataTables warning: table id=ords-table - Ajax error. For more information about this error, please see http://datatables.net/tn/7"

I have a one to one relationship with two models and tables: Ord with the ords table (orders) and Rone  with the rones table (projects).

[...]

PR #585, opened 2016-05-31
HasOne relation uses different methods to get foreign and other key

peterquentin:
This should fix any issues one might have when using a HasOne relation.
The API to get a foreign and other key are different then other relation types.

I believe it should also fix issue #567


yajra:
@peterquentin PR works like a charm. Merged and released on v6.11.5. Thanks a lot! 👍

PR #597, opened 2016-06-08
Added Auto-Index Column.

cirdog:
An Index-Column is useful when users need to reference to what they see
on screen while working with others.

### DIFF
--- a/src/Engines/BaseEngine.php
+++ b/src/Engines/BaseEngine.php
@@ -669,7 +669,8 @@ abstract class BaseEngine implements DataTableEngineContract
         $processor = new DataProcessor(
             $this->results(),
             $this->columnDef,
-            $this->templates
+            $this->templates,
+            $this->request['start']
         );
 
         return $processor->process($object);
--- a/src/Html/Builder.php
+++ b/src/Html/Builder.php
@@ -405,6 +405,31 @@ class Builder
     }
 
     /**
+     * Add a index column.
+     *
+     * @param  array $attributes
+     * @return $this
+     */
+    public function addIndex(array $attributes = [])
+    {
+        $attributes = array_merge([
+            'defaultContent' => '',
+            'data'           => 'index',
+            'name'           => 'index',
+            'title'          => '',
+            'render'         => null,
+            'orderable'      => false,
+            'searchable'     => false,
+            'exportable'     => false,
+            'printable'      => true,
+            'footer'         => '',
+        ], $attributes);
+        $this->collection->push(new Column($attributes));
+
+        return $this;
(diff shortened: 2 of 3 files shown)


cirdog:
Sorry, my 2nd commit "Merge remote-tracking.... blah blah" was a mistake, please kindly ignore it.

I haven't yet mastered the steps of issuing a PR, always attaching junks onto the first commit.

Only the 1st commit is intended, sorry again.

yajra:
@cirdog thanks for PR. Will try to review and test this as soon as I can! 👍

yajra:
Just tested this and seems to work fine. 

Usage notes for future reference:
1. On html builder, use `$builder->addIndex()`.
2. To render/include on output, you need to use `$dataTable->addColumn('index', '')`.

On the other hand, I think this may be a potential breaking change because an `index` column might be used on a table? I suggest we change the index name to something that might not conflict a real column name? Maybe use something like `dt_row_index`?

On the engine side, maybe we can also add something like `addIndexColumn()` instead of using `addColumn` with empty string param?

Thanks!

cirdog:
All good ideas! Please go ahead.

But I'd like to add that the original "action" column name would also conflict with an actual DB field. So maybe we standardize them into dt_row_index & dt_row_action?

Although I personally would prefer "_index" & "_action" since by convention an underscore prefix should be rare & intentional enough to avoid name conflicts.

But it's really up to your preference!

yajra:
Good point for the action field but I think it's a bit late to change it now since most of the users may have adapted using it. `_index` is fine for me too but come to think of it, the plugin already have something like [DT_RowId](https://datatables.net/examples/server_side/ids.html) so I think it would make sense if we use something like `DT_RowIndex` like on this [client side index](https://datatables.net/reference/api/row().index())?

cirdog:
I'm really okay with your naming scheme, please go ahead.

As for a future-proof scheme, maybe later these names could be user-configurable? 

Then backward-compatibility can be preserved for users who prefer to stick with names that they know that are conflict-free, while other users can also safely use "action" as a DB field name without any worry.

User-configurable special column names can also facilitate a smoother upgrade path, to safely phase out generic names like "action" for a more robust package.

yajra:
Sorry for the delays. PR merged and released on v6.13.0. 

I also implemented what we have discussed here and use `DT_Row_Index` as the key for the index but is configurable via `datatables.index_column` key. I also added `addIndexColumn()` method which can be used with/without the builder. Thanks!

```php
return $this->datatables
            ->collection($this->query())
            ->addIndexColumn()
            ->make(true);
```

Issue #578, opened 2016-05-26
Speeding up or disabling entries count


marijnhurkens:
I really like this package and I am using it in several projects, but I have a problem with the speed of the queries it performs.

My situation is as follows:

I am retrieving my information from a table called Students while also joining 2 tables called Users and Addresses. I also have some extra filters next to the global search. Using the filter option of this package disables the global search, so I am filtering manually before sending the query to Datatables::of($query).

This all works great, but the entries count query is taking a 150 ms to complete, while the main select query is taking 6 ms. The database has about a 1000 entries and is expected to grow to 10,000+.

My code:
```php
$students = Student::join('users', function ($join) {
            $join->on('users.owner_id', '=', 'students.id');
            $join->on('users.owner_type', '=', 'App\Models\Student');
        })
            ->join('addresses', function ($join) {
                $join->on('addresses.owner_type', '=', 'App\Models\Student');
                $join->on('addresses.owner_id', '=', 'students.id');
            });

        if ($request->has('city')) {
            $students->where('addresses.city',  $request->get('city'));
        }

        if ($request->has('status')) {
            $students->where('students.status',  $request->get('status'));
        }
        
        $students->select('students.created_at', 'students.first_name', 'students.last_name', 'users.email',
            'students.telephone_number', 'students.gender', 'students.status', 'students.id', 'addresses.city');
        
        return Datatables::of($students)
            ->editColumn('created_at', '{!! $created_at->toDateString() !!}')
            ->make(TRUE);
```

The resulting query for the entries count is (157.07 ms):
```sql
select count(*) as aggregate from (select '1' as "row_count" from "students" 
inner join "users" on "users"."owner_id" = "students"."id" and "users"."owner_type" = "App\Models\Student" 
inner join "addresses" on "addresses"."owner_type" = "App\Models\Student" and "addresses"."owner_id" = "students"."id" 
where "students"."deleted_at" is null) count_row_table
```

And the main select query (5.98 ms):

```sql
select "students"."created_at", "students"."first_name", "students"."last_name", "users"."email", "students"."telephone_number", "students"."gender", "students"."status", "students"."id", "addresses"."city" 
from "students" inner join "users" on "users"."owner_id" = "students"."id" and "users"."owner_type" = "App\Models\Student" 
inner join "addresses" on "addresses"."owner_type" = "App\Models\Student" and "addresses"."owner_id" = "students"."id" 
[...]

PR #621, opened 2016-07-01
Add count parameter to make()

TobyRyuk:
Added a $count parameter as the third argument to make(). This will let you count the records yourself to make sure the count query is as minimal as possible, especially when you have complicated queries. As this is nullable, everything works as it previously did.

### DIFF
--- a/src/Contracts/DataTableEngineContract.php
+++ b/src/Contracts/DataTableEngineContract.php
@@ -73,7 +73,8 @@ interface DataTableEngineContract
      *
      * @param bool $mDataSupport
      * @param bool $orderFirst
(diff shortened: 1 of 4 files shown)


yajra:
@TobyRyuk, thanks for the PR but I already did something like this a week ago. Please see this issue #578 and let me know if it works for you.

TobyRyuk:
Oh, totally missed that. Sorry for the unnecessary PR.

yajra:
No problem and thanks for the intent to contribute. I also added a demo for ref: https://datatables.yajrabox.com/eloquent/manual-count

PR #649, opened 2016-07-18
Transformer instance is kept, otherwise new instance is created.

ssipos90:
In my use case, the transformer constructor that I provide has arguments which are lost because of re-instancing the transformer.
This patch allows for creating a transformer instance if the class name is provided, otherwise, if an instance is provided, it keeps it.

### DIFF
--- a/src/Engines/BaseEngine.php
+++ b/src/Engines/BaseEngine.php
@@ -648,11 +648,11 @@ abstract class BaseEngine implements DataTableEngineContract
             //If parameter is class assuming it requires object
             //Else just pass array by default
             if ($parameter->getClass()) {
-                $resource = new Collection($this->results(), new $this->transformer());
+                $resource = new Collection($this->results(), $this->createTransformer());
             } else {
                 $resource = new Collection(
                     $this->getProcessedData($object),
-                    new $this->transformer()
+                    $this->createTransformer()
                 );
             }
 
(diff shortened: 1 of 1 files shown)


ssipos90:
If you accept this, please be a good fella and tag a new version so I can use your repo. Thanks!

src/Engines/BaseEngine.php
```diff
@@ -670,6 +670,20 @@ public function render($object = false)
...
 
     /**
+     * Get or create transformer instance
+     *
+     * @return \League\Fractal\TransformerAbstract
+     */
+    protected function createTransformer()
+    {
+        if ($this->transformer instanceof \League\Fractal\TransformerAbstract) {
+            return $this->transformer;
```
yajra: I think there is a missing `new` here?

```php
return new $this->transformer;
```

yajra:
@ssipos90, I think this line will always return true since all transformer implements it?

```php
if ($this->transformer instanceof \League\Fractal\TransformerAbstract) {
```

Can you please provide some snippet on how I can replicate the issue being fixed here? Thanks!

ssipos90:
It depends on how you use the setter.
$engine->setTransformer(new BelovedTransformer()); will go through the if.
$engine->setTransformer(BelovedTransformer::class); won't.

ssipos90:
This makes use of the fact that you didn't typehint the TransformerAbstract on the setter.

yajra:
Merged and released on v6.16.0, thanks!

ssipos90:
Thanks!

Issue #348, opened 2016-01-19
[7.0] Add support for DataTable Service via POST ajax request.


yajra:
In relation to issue #341 discussions.

nikunjkabariya:
Have you changed/implemented this post method changes?
When will you implement POST method as its not working with Export options(pdf, xls, csv)?
Currently GET method is used. When i change it to POST, so for export option page stays blank.
Thanks

yajra:
@nikunjkabariya this is not yet implemented. If you can, PR's are welcome. Thanks!

vesper8:
really could use this.. it's causing me a lot of pain :(

gadius:
@mastrip2 Just nailed it; changed my initialization as follow.

var tableIndex=$('#tableIndex').DataTable({
"order": [[ 0, "asc" ]],
"aLengthMenu": [[5, 10, 25], [5, 10, 25]],
"iDisplayLength": 10,
"dateFormat": 'yyyy-mm-dd',
"processing": true,
"serverSide": true,
**"ajax": {
'url': '{{route("clinics.api.laboratoryIndex")}}',
'type': 'POST',
'headers': {
'X-CSRF-TOKEN': '{{ csrf_token() }}'
}
},**
"columns":[
{data: 'products', name: 'products'},

],
"columnDefs": [
{ "searchable": false, "bSortable": false, "targets": 0 }
],
language: {
"sProcessing": "Procesando...",
"sLengthMenu": "Mostrar MENU registros",
"sZeroRecords": "No se encontraron resultados",
"sEmptyTable": "Ningún dato disponible en esta tabla",
"sInfo": "Mostrando registros del START al END de un total de TOTAL registros",
"sInfoEmpty": "Mostrando registros del 0 al 0 de un total de 0 registros",
"sInfoFiltered": "(filtrado de un total de MAX registros)",
"sInfoPostFix": "",
"sSearch": "Buscar: ",
"sUrl": "",
"sInfoThousands": ",",
"sLoadingRecords": "Cargando...",
"oPaginate": {
"sFirst": "Primero",
"sLast": "Último",
"sNext": "Siguiente",
"sPrevious": "Anterior"
},
"oAria": {
"sSortAscending": ": Activar para ordenar la columna de manera ascendente",
"sSortDescending": ": Activar para ordenar la columna de manera descendente"
}
}
});

vesper8:
Using the default implementation of this package "as a service", can you explain where you made these modifications?

[...]

PR #671, opened 2016-07-29
use post request for export and print buttons

alex-LE:
I've changed the behaviour of the export and print buttons to use a post instead of get request.
I got some issues with long urls as i have to many filters and columns in my datatable.

As jquery post does not support downloading files, i've used a xhr function i've found [here](http://stackoverflow.com/questions/16086162/handle-file-download-from-ajax-post).

Laravel requires a csrf token for post requests, so i've added that to the script.blade.php file (i'm not sure if this is the best solution).

I've tested the excel, csv, pdf and print button in Firefox and Google Chrome.

I had to change my routes file also to handle the get and post request to the datatable:

`Route::match(['get', 'post'], '/', function () { ... });`

Fixes issue #348

### DIFF
--- a/src/resources/assets/buttons.server-side.js
+++ b/src/resources/assets/buttons.server-side.js
@@ -1,12 +1,60 @@
 (function ($, DataTable) {
     "use strict";
 
-    var _buildUrl = function(dt, action) {
-        var url = dt.ajax.url() || '';
+    var _buildParams = function(dt, action) {
         var params = dt.ajax.params();
         params.action = action;
+        params.columns = addVisiblePropertyToColumns(params.columns, dt);
+        params._token = $.fn.dataTable.defaults.csrf_token;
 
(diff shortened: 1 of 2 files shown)


yajra:
@alex-LE thanks for this great PR. I think it would be better if we can adopt to what method the DT ajax used? If `get` method was used, we stick to `get` for export and same goes for `post`. This way, it would not break any installations.

alex-LE:
Makes sense, i will be back in the office next week.
I will have a look at it.

yajra:
@alex-LE thanks for this PR. Will be implementing this feature on v7 by adding a new button like `excelPost`, `csvPost` & `pdfPost`. This way, it would not break any existing installations and we will have more options too. My friend started this approach on this PR #826 and will continue the development from there. Thanks!

Issue #702, opened 2016-08-22
SetTotalRecords in CollectionEngine


Nks:
### Summary of problem or feature request

Using collection we can't rewrite recordsTotal with ```setTotalRecords()``` method because it's not using in the CollectionEngine.php. 

### Code snippet of problem

Trying to do something like this:
```
return Datatables::of(new Collection($result))
                         ->setTotalRecords(100000)
                         ->make(true);
```
And it always return me only 10 (this is how much items we have in collection).
In [CollectionEngine.php](https://github.com/yajra/laravel-datatables/blob/6.0/src/Engines/CollectionEngine.php#L104) it calling method `count()` from Collection method. It seems we can add here ability to rewrite count. In this case we can get data from somewhere with limits and offset and rewrite totalRecords with it.

### System details

- Ubuntu 14.04.4 LTS
- PHP 7.0.3-13+deb.sury.org~trusty+1 (cli) ( NTS )
- Laravel 5.2.43
- Laravel-Datatables 6.17.0

yajra:
Fix on v6.17.1, thanks for reporting!

Nks:
@yajra many thanks for updates!

PR #703, opened 2016-08-22
SetTotalRecords in CollectionEngine

Nks:
Issue described in #702 
For now we can use SetTotalRecords when we using CollectionEngine.
Example:
```
        $offset = Input::get('start', null);
        $limit = Input::get('length', null);
        $dataSourceRows = DataSourceRow::whereDataSourceId($dataSourceId);
        $dataSourceRowsCount = $dataSourceRows->count() / $dataSourceColumnsCount;
        if (!is_null($offset) && !is_null($limit)) {
            $dataSourceRows->limit($limit * $dataSourceColumnsCount)
                           ->offset($offset * $dataSourceColumnsCount);
        }

        $rows = $dataSourceRows->orderBy('id')->get();

        foreach ($rows as $row) {
            $result[$row->row_id][$row->dataSourceColumns->title] = $row->value;
        }
        return Datatables::of(new Collection($result))
                         ->setTotalRecords($dataSourceRowsCount)
                         ->make(true);
```
We are retrieving data from model with a lot of items and with limits, doing some stuff with it and creating Collection with this data. Also we setting up totalRecords by manual with `setTotalRecords($dataSourceRowsCount)` and it retrieving us correct json data for the datatables.

### DIFF
--- a/src/Engines/BaseEngine.php
+++ b/src/Engines/BaseEngine.php
@@ -186,9 +186,7 @@ abstract class BaseEngine implements DataTableEngineContract
      * @var array
      */
     private $appends = [];
-
-
-
+    
     /**
      * Setup search keyword.
      *
--- a/src/Engines/BaseEngine.php
+++ b/src/Engines/BaseEngine.php
@@ -174,11 +174,20 @@ abstract class BaseEngine implements DataTableEngineContract
     protected $orderCallback;
 
     /**
+     * Custom rewriting totals
+     *
+     * @var bool
+     */
+    protected $rewriteTotals = false;
+
+    /**
      * Array of data to append on json response.
      *
      * @var array
      */
     private $appends = [];
+
+
 
     /**
      * Setup search keyword.
@@ -960,6 +969,8 @@ abstract class BaseEngine implements DataTableEngineContract
      */
     public function setTotalRecords($total)
     {
+        $this->rewriteTotals = true;
+
         $this->totalRecords = $total;
 
         return $this;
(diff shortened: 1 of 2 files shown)


yajra:
@Nks thanks for the PR but I think the fix should be on the total count? No need for rewrite flag?

Can you please check if the snippet below will work for you? 

I forgot to implement this feature on collection engine, thanks for the ping!

```php
public function totalCount()
{
    return $this->totalRecords ? $this->totalRecords : $this->count();
}
```

yajra:
Implemented on #704. Released on v6.17.1, thanks!

src/Engines/CollectionEngine.php
```diff
@@ -205,10 +205,12 @@ function ($row) use ($column, $keyword) {
      */
     public function paging()
     {
-        $this->collection = $this->collection->slice(
-            $this->request['start'],
-            (int) $this->request['length'] > 0 ? $this->request['length'] : 10
-        );
+        if(!$this->rewriteTotals) {
```
Nks: Hey @yajra 
I think we need this in the live code.
Look at #706 for more information

src/Engines/BaseEngine.php
```diff
@@ -960,6 +967,8 @@ protected function isOracleSql()
      */
     public function setTotalRecords($total)
     {
+        $this->rewriteTotals = true;
```
Nks: Also this we can use in other logic when we not need paging, for example.

Issue #706, opened 2016-08-23
Issue with paging with setTotalRecords in Collection


Nks:
### Summary of problem or feature request

In [CollectionEngine](https://github.com/yajra/laravel-datatables/blob/6.0/src/Engines/CollectionEngine.php#L208) we are using `slice()` method in paging method. When system getting data from Collection it slicing result array with start and length and if we are using `setTotalRecords` it can occur error and not returning any data on any page instead first. 
It was resolved by my PR #703. See comments in the commit.

Nks:
![image](https://cloud.githubusercontent.com/assets/349293/17884889/50d86750-6923-11e6-9452-db466c3afe84.png)
Here it is result from 
```Datatables::of(new Collection($result))
                         ->setTotalRecords($dataSourceRowsCount)
                         ->make(true);```
In `$result` we have only 10 records and when we using paging and skiping first 10 records (moving on 2nd page) it slicing data and retrieving nothing.

PR #707, opened 2016-08-23
Issue #706: Fix paging

Nks:
Issue described in #706

### DIFF
--- a/src/Engines/BaseEngine.php
+++ b/src/Engines/BaseEngine.php
@@ -174,11 +174,11 @@ abstract class BaseEngine implements DataTableEngineContract
     protected $orderCallback;
 
     /**
-     * Custom rewriting totals
+     * Skip paginate when it need.
      *
      * @var bool
      */
-    protected $rewriteTotals = false;
+    protected $skipPaging = false;
 
     /**
      * Array of data to append on json response.
@@ -636,7 +636,7 @@ abstract class BaseEngine implements DataTableEngineContract
      */
     public function paginate()
     {
-        if ($this->request->isPaginationable()) {
+        if ($this->request->isPaginationable() && !$this->skipPaging) {
             $this->paging();
         }
     }
@@ -967,9 +967,16 @@ abstract class BaseEngine implements DataTableEngineContract
      */
     public function setTotalRecords($total)
     {
-        $this->rewriteTotals = true;
-
         $this->totalRecords = $total;
+
+        return $this;
+    }
+
+    /**
+     * Skipping paging when we need
+     */
+    public function skipPaging(){
+        $this->skipPaging = true;
 
         return $this;
     }
--- a/src/Engines/BaseEngine.php
+++ b/src/Engines/BaseEngine.php
@@ -174,6 +174,13 @@ abstract class BaseEngine implements DataTableEngineContract
     protected $orderCallback;
 
     /**
+     * Custom rewriting totals
+     *
+     * @var bool
+     */
+    protected $rewriteTotals = false;
+
+    /**
      * Array of data to append on json response.
      *
      * @var array
@@ -960,6 +967,8 @@ abstract class BaseEngine implements DataTableEngineContract
      */
     public function setTotalRecords($total)
     {
+        $this->rewriteTotals = true;
+
         $this->totalRecords = $total;
 
         return $this;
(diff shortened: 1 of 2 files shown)


yajra:
I just added a simple demo for this and seems to work fine with me.
https://datatables.yajrabox.com/collection/total-records

Using your PR, it will skip the pagination.

Nks:
Try it with array which have only 10 elements but when you need setup totalRecords by manual.
For example:
$users = new Collection;
        for ($i = 0; $i < 10; $i++) {
            $users->push([
                'id'         => $i + 1,
                'name'       => md5(microtime()),
            ]);
        }

        return Datatables::of($users)->->setTotalRecords(20)->make(true);
```

Перевести

Идея заключается в том, чтобы использовать limit and offset при получении данных откуда либо и после этого использовать их при выводе. 
Например, есть таблица с 2000000 записей. В вашем случае мы получим все данные из базы (все 2000000 записей), что существенно замедлит ответ. Я же предлагаю использовать limit и offset для получения данных откуда либо и после этого использовать эти данные для вывода.

The idea is to use a limit and offset when receiving data from any and then use them in the output.
For example, there is a table with 2000000 records. In your case, we'll get all the data from the database (all records 2000000), which would slow down the response. I suggest using limit and offset to obtain data from any and then use that data to display.

Example which will works fine with my PR:
```
$dataSourceColumnsCount = $dataSource->dataSourceColumns->count();

        $result = [];
        $offset = Input::get('start', null);
        $limit = Input::get('length', null);

        $dataSourceRows = DataSourceRow::whereDataSourceId($dataSourceId);
        $dataSourceRowsCount = $dataSourceRows->count() / $dataSourceColumnsCount; //counting all elements
        if (!is_null($offset) && !is_null($limit)) {
//retrieving all data from database with limits and offset
            $dataSourceRows->limit($limit * $dataSourceColumnsCount)
                           ->offset($offset * $dataSourceColumnsCount); 
        }

        $rows = $dataSourceRows->orderBy('id')->get();

        /**
         * @var $row DataSourceRow
         */
//preparing data for the datatables
        foreach ($rows as $row) {
            $result[$row->row_id][$row->dataSourceColumns->title] = $row->value;
        }



        return Datatables::of(new Collection($result))
//setting up total records because in $result we have only limited data
                         ->setTotalRecords($dataSourceRowsCount)
                         ->make(true);
```

It's clear?

**Comment by Nks** [comment]
BTW we can use `skipPaging()` method or something like this. What do you think?

**Comment by yajra** [comment]
So the issue is when you have smaller total than the actual limited results?

Maybe you should add a condition then:

```php
$users = new Collection;
for ($i = 0; $i < 10; $i++) {
    $users->push([
        'id' => $i + 1,
        'name' => md5(microtime()),
    ]);
}

$total = 20;

return Datatables::of($users)
    ->setTotalRecords(count($users) > $total ? $total : count($users))
    ->make(true);
```

yajra:
I also pushed another fix to not allow filteredTotal to be larger than the total count. Is this what you need? Please see https://github.com/yajra/laravel-datatables/pull/708 if it makes sense. Thanks!

Nks:
No, it's not work. 
Look:
we have 10 elements in the collection and total = 20. We trying to get next 10 elements (start = 10 and length = 10) and using it in the limit and in the offset. From database we will get only 10 elements and `slice()` method will slice it from start (from 10 element) to length (to 20 element) but we *have not* this elements. In the collection we still have only 10 elements and it's not correct. Of course if we have full collection with full elements we not needed change total and all works fine, but if we want limit source (to avoid load on the database with a lot of data) it not works fine.

Nks:
In current version only 1 not good solution: remove offset from source query and multiply limit and offset but the further we move through the pages number of items in the collection will grow. It's not good.

yajra:
@Nks ok so you already implemented the paging when fetching the collection hence DT paging should be disabled? Then I guess, `skipPaging()` as you suggested would make sense. Can you please submit a PR if you can? You current PR will always skip paging. Will look at this again tomorrow. Thanks!

Nks:
@yajra sure, I too this it will be better than current. Thanks!

Nks:
@yajra skipPaging() method implemented. Please take a look and if all ok apply it asap ;)

Issue #732, opened 2016-09-04
Column search for eager loaded HasMany-relationship


SRautila:
### Summary of problem or feature request

I was trying to get column search working for my table with a column based on a eager loaded HasMany-relationsship and got the following error.

`BadMethodCallException in Builder.php line 2435:
Call to undefined method Illuminate\Database\Query\Builder::getQualifiedForeignKey()`

`9. at HasMany->getQualifiedForeignKey() in QueryBuilderEngine.php line 435` 

Looking at the code in `QueryBuilderEngine.php` it seems that all that's needed for it to work is to treat it the same way as for HasOne as [HasMany also has the getForeignKey()](https://laravel.com/api/5.3/Illuminate/Database/Eloquent/Relations/HasMany.html#method_getForeignKey)-method and not getQualifiedForeignKey().

So after changing `($model instanceof HasOne)` to `($model instanceof HasOne || $model instanceof HasMany)` column-search works fine for my HasMany-column. I submitted at PR #733 with the change. I hope I did everything correctly as I'm new to contributing to open source. :)

### My code which didn't work before the fix

In the controller:

```php
$customers = Customer::with([ 'phone_numbers' => function ($query) {
                $query->orderBy('type', 'desc');
            }])->select('customers.*');
return Datatables::of($customers)->make(true)
```

Column definition:

```js

{data: 'phone_numbers', name: 'phone_numbers.number', orderable: false, render: function (data) {
                        var numbers = [];
                        data.forEach(function (datum) {
                            numbers.push(datum.number);
                        });
                        return numbers.join(", ");
                    }}
```

### System details

- Operating System Windows 10
- PHP Version 5.6.15
- Laravel Version 5.3.6
- Laravel-Datatables Version 6.18.0

yajra:
Should be fixed on v6.18.1. Thanks for the PR!

yajra:
Demo added for HasMany. Thanks!
https://datatables.yajrabox.com/relation/has-many

zipsoft:
Hi Yajra I can't access HasMany demo page:

This site can’t be reached datatables.yajrabox.com’s server IP address could not be found.

Do you have any alternative links?

PR #733, opened 2016-09-04
Fix to allow join of eager loaded column based on a HasMany-relationship

SRautila:
Hi

I was having issues implementing column search for a column showing data from a eager loaded HasMany-relationship. Treating HasMany in the same way as HasOne fixed it for me as HasMany has the getForeignKey()-method as well.

See the issue i created: #732


yajra:
Merged and released on v6.18.1. Thanks a lot! 👍

ssipos90:
Those two classes extend a base class:
`if ($model instanceof HasOneOrMany) {`

yajra:
@ssipos90 good point, didn't notice that. Is it like that since Laravel 5.0? Can you send a PR to refactor it for proper credits on contributions? Thanks!

ssipos90:
Since 4.x

ssipos90:
No credits needed, you do it, please! :D

PR #735, opened 2016-09-05
Fix to Str::lower() on array

jaketoolson:
With eager loaded relationships on a collection, searching a column containing relationships fails due to `Str::lower()` expecting a string, but an array is being passed.  Converting array to json_decoded string works.

### Code snippet of problem
`Str::lower($value)`

Fix:
`$value = is_array($value) ? json_decode(json_encode($value), true) : $value;`

### System details

- Operating System: Ubuntu
- PHP Version: 7.0.10
- Laravel Version: Laravel 5.2.x
- Laravel-Datatables Version: 6.18.0

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,7 +19,7 @@
         "league/fractal": "~0.12",
         "laravelcollective/html": "~5.0",
         "maatwebsite/excel": "^2.0",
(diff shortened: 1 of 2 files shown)


jaketoolson:
Let me submit a new one -- new bugs found.

PR #738, opened 2016-09-07
Updated league/fractal version to fix conflicts

tyloo:
We just simply update league/fractal version

Signed-off-by: Julien Bonvarlet <julien.bonvarlet@carvinglabs.com>

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -16,7 +16,7 @@
         "illuminate/view": "~5.0",
         "illuminate/http": "~5.0",
         "illuminate/filesystem": "~5.0",
-        "league/fractal": "~0.12",
+        "league/fractal": "~0.14",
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -1,5 +1,5 @@
 {
-    "name": "yajra/laravel-datatables-oracle",
+    "name": "carvinglabs/laravel-datatables-oracle",
```
yajra: Changing the name might break the installations. Will implement this on a new PR with credits to you. Thanks!

yajra:
I think it's better that you update your PR than I fixing the composer name so that your contribution will be properly added on contributors list. Will hold off my own PR for a week. Thanks a lot and sorry for delayed response.

PR #765, opened 2016-09-23
Support regex column searches when using CollectionEngine

cadenzajon:
The QueryBuilderEngine has support for using Oracle queries to search for regular expressions when a columns 'regex' flag is enabled. This update adds similar functionality to the CollectionEngine using PHP's built-in preg functions.

### DIFF
--- a/src/Engines/CollectionEngine.php
+++ b/src/Engines/CollectionEngine.php
@@ -189,13 +189,13 @@ class CollectionEngine extends BaseEngine
                         $value = Arr::get($data, $column);
 
                         if ($this->isCaseInsensitive()) {
-                            if($regex) {
-                                return preg_match('/' . Str::lower($keyword) . '/', Str::lower($value)) == 1;
+                            if ($regex) {
+                                return preg_match('/' . $keyword . '/i', $value) == 1;
                             } else {
                                 return strpos(Str::lower($value), Str::lower($keyword)) !== false;
                             }
                         } else {
-                            if($regex) {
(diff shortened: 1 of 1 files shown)


src/Engines/CollectionEngine.php
```diff
@@ -177,20 +177,29 @@ public function columnSearch()
...
                 $this->collection = $this->collection->filter(
-                    function ($row) use ($column, $keyword) {
+                    function ($row) use ($column, $keyword, $regex) {
                         $data = $this->serialize($row);
 
                         $value = Arr::get($data, $column);
 
                         if ($this->isCaseInsensitive()) {
-                            return strpos(Str::lower($value), Str::lower($keyword)) !== false;
+                            if($regex) {
```
yajra: Please add a space after `if ()`. Thanks!

src/Engines/CollectionEngine.php
```diff
@@ -177,20 +177,29 @@ public function columnSearch()
...
+                                return preg_match('/' . Str::lower($keyword) . '/', Str::lower($value)) == 1;
```
yajra: I think we can use preg_match with `i` flag as per this [SO post](http://stackoverflow.com/questions/12411037/how-do-i-make-this-preg-match-case-insensitive)?

cadenzajon:
Thanks, I have made those updates!

yajra:
Released on v6.19.0, thanks a lot! 👍

PR #786, opened 2016-10-11
Support of using soft deletes trait in models.

DDiimmkkaass:
This is required when a user uses a model with soft deletes trait (use Illuminate\Database\Eloquent\SoftDeletes), but does not include to select trashed items (withTrashed()), then the number of selected elements and elements in the pagination begins to differ.
Commit adds the ability to manage it and select or not deleted records in count function

### DIFF
--- a/src/Engines/QueryBuilderEngine.php
+++ b/src/Engines/QueryBuilderEngine.php
@@ -407,7 +407,11 @@ class QueryBuilderEngine extends BaseEngine
      */
     private function modelUseSoftDeletes()
     {
-        return in_array('Illuminate\Database\Eloquent\SoftDeletes', class_uses($this->query->getModel()));
+        if ($this->query_type == 'eloquent') {
+            return in_array('Illuminate\Database\Eloquent\SoftDeletes', class_uses($this->query->getModel()));
+        }
+        
+        return false;
     }
     
     /**
--- a/src/Engines/QueryBuilderEngine.php
+++ b/src/Engines/QueryBuilderEngine.php
@@ -100,6 +100,11 @@ class QueryBuilderEngine extends BaseEngine
             $row_count = $this->wrap('row_count');
             $myQuery->select($this->connection->raw("'1' as {$row_count}"));
         }
+    
+        // check for select soft deleted records
+        if (!$this->withTrashed && $this->modelUseSoftDeletes()) {
+            $myQuery->whereNull($myQuery->getModel()->getTable().'.deleted_at');
+        }
 
         return $this->connection->table($this->connection->raw('(' . $myQuery->toSql() . ') count_row_table'))
                                 ->setBindings($myQuery->getBindings())->count();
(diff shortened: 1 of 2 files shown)


src/Engines/QueryBuilderEngine.php
```diff
@@ -394,7 +399,28 @@ private function getSearchKeyword($i, $raw = false)
...
+     * @return boolean
+     */
+    private function modelUseSoftDeletes()
+    {
+        $app = app();
+        
+        //check for laravel version
+        //in 4 and 5 versions SoftDeletes trait has a different names
+        //we can check only first number
+        if (strpos($app::VERSION, '5') === 0) {
```
yajra: Maybe this check is not needed since v6.0 only support Laravel 5++?

src/Engines/QueryBuilderEngine.php
```diff
@@ -394,7 +399,28 @@ private function getSearchKeyword($i, $raw = false)
...
+            $class = 'Illuminate\Database\Eloquent\SoftDeletes';
+        } else {
+            $class = 'Illuminate\Database\Eloquent\SoftDeletingTrait';
+        }
+        
+        return in_array($class, class_uses($this->query->getModel()));
```
yajra: I think we should check first if the instance is a model like:
```php
if ($this->query_type == 'eloquent') {
    return $this->query->getModel() instanceof 'Illuminate\Database\Eloquent\SoftDeletes';
}

return false;

File: src/Engines/QueryBuilderEngine.php [hunk 3]
```diff
@@ -100,6 +100,11 @@ public function count()
             $row_count = $this->wrap('row_count');
             $myQuery->select($this->connection->raw("'1' as {$row_count}"));
         }
+    
+        // check for select soft deleted records
+        if (!$this->withTrashed && $this->modelUseSoftDeletes()) {
```
yajra: Please add a space after negation `if (! $this...)`. Thanks!

src/Engines/QueryBuilderEngine.php
yajra: Removing this check might also fix the unit test. Thanks!

src/Engines/QueryBuilderEngine.php
DDiimmkkaass: ok, thanks.

src/Engines/QueryBuilderEngine.php
DDiimmkkaass: I not sure, that i understand your comment aright, but model is not a instance of SoftDeletes.
SoftDeletes - is only a trait, that uses in model and this check will be not correct.

yajra: > I not sure, that i understand your comment aright, but model is not a instance of SoftDeletes.

Oh right, I think that is not the right way to check a trait? Not sure, haven't tried check a trait yet ^_^

DDiimmkkaass: In fact, this method(class_uses) is created for this(check of triat usage). I try to find other methods, but in reality it is not so much. Anoter way is check of method_exists(), but i think this is worse. And one more way, do this, is use of interfaces(contracts) but for the package is not suitable. 
The only thing that we can fix it use instans of class SoftDeletes
in_array(new SoftDeletes, class_uses($this->query->getModel()))
but it will create a new object, which essentially would only take a memory

yajra:
@DDiimmkkaass thanks a lot! Released on v6.20.0 💯

yajra:
BTW, I also added a demo for your PR. https://datatables.yajrabox.com/eloquent/with-trashed as a proof of concept. Thanks!

DDiimmkkaass:
Cool, thanks!

michail1982:
@DDiimmkkaass , @yajra 
This feature does not respect
```php 
const DELETED_AT
```
and does not make sence in eloquent mode

PR #804, opened 2016-10-20
support to morphToMany

pcastrotigre:
Adding support for morphToMany to search

### DIFF
--- a/src/Engines/QueryBuilderEngine.php
+++ b/src/Engines/QueryBuilderEngine.php
@@ -3,6 +3,7 @@
 namespace Yajra\Datatables\Engines;
 
 use Closure;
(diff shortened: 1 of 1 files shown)


yajra:
@pcastrotigre thanks for the support. Will try to review/test this asap. 👍

pcastrotigre:
It works only for the Search. However as the Sort is so subjective, I only skipped so by default, _**Sort will do nothing if the orderable column is from a morphToMany relationship**_. 

I would suggest that the morphToMany column should be set to `orderable=false` in the javascript or define a `callback` to sort by it.

yajra:
Released on v6.21.0. Thanks a lot! 💯

yajra:
Demo added for this use case. Thanks!

https://datatables.yajrabox.com/relation/morph-to-many

PR #826, opened 2016-11-11
[7.0] Add post button for exporting data to fix compatibility with IE.

gredimano:
Add export post buttons for Excel, CSV, PDF data. 
Supported Browsers: IE, Chrome, Firefox.

Please check laravel-datables demo for quick example :).
https://datatables.yajrabox.com/

Thanks for this great PR. @alex-LE
https://github.com/yajra/laravel-datatables/pull/671

### DIFF
--- a/src/resources/assets/buttons.server-side.js
+++ b/src/resources/assets/buttons.server-side.js
@@ -1,7 +1,66 @@
 (function ($, DataTable) {
     "use strict";
 
-    var _buildUrl = function(dt, action) {
+    var _buildParams = function (dt, action) {
+        var params = dt.ajax.params();
+        params.action = action;
+        params._token = $.fn.dataTable.defaults.csrf_token;
+
+        return params;
+    };
+
+    var _downloadFromUrl = function (url, params) {
+        var postUrl = url + '/export';
+        var xhr = new XMLHttpRequest();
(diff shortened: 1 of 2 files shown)


yajra:
Can you please fix the unit tests. Thanks!

TonyRyan:
What's the status on this?

underdpt:
The tests are failing because of the csrf addition to `scripts.blade.php`. This addition is only neccesary if using post, and cannot be moved to `buttons.server-side.js´ because we cannot retrieve the csfr token from a javascript file, it has to be included on a blade template.

So I think we can:
- Update the failing tests to add the csfr token to the expected output
- Remove all the csrf code and simply document that.

I would go for the second option because:
- There is a POST way of sending the search parameter and there's no csfr mention at all. It's up to the developer to deal with it (for example, I'm adding it to the headers via `$.ajaxSetup()`)
- It may not be neccesary at all. You can disable csfr protection on some routes, or on all routes modifying the `VerifyCsrfToken` middleware.

So maybe the best is to simply document that and let the developer decide how to deal with csrf protection.

Regards,

yajra:
Fixed via PR https://github.com/yajra/laravel-datatables-html/issues/13.

PR #833, opened 2016-11-16
Fix html builder instance, close #830

### LINKED ISSUE #830: Should the html builder be cached?
For DataTables Service usage, each call to `Services\DataTable::builder` or `Services\DataTable::html` will return a new Html Builder instance. Should we cache this instance for further use?

---

The current source code in `Yajra\Datatables\Services\DataTable`:

```php
/**
 * Optional method if you want to use html builder.
 *
 * @return \Yajra\Datatables\Html\Builder
 */
public function html()
{
    return $this->builder();
}

/**
 * Get Datatables Html Builder instance.
 *
 * @return \Yajra\Datatables\Html\Builder
 */
public function builder()
{
    return $this->datatables->getHtmlBuilder();
}
```

And the `Datatables::getHtmlBuilder`:

```php
public function getHtmlBuilder()
{
    return app('Yajra\Datatables\Html\Builder');
}
```

---

Maybe the html builder could be cached using a variable, then each call to `html` will return the same instance.

```php
/**
 * Get Datatables Html Builder instance.
 *
 * @return \Yajra\Datatables\Html\Builder
 */
public function builder()
{
    if (is_null($this->builder)) {
        $this->builder = $this->datatables->getHtmlBuilder();
    }

    return $this->builder;
}
```

ElfSundae:


yajra:
👍  Thanks!

PR #849, opened 2016-11-22
Patch for fix ambiguous field error for relations.

ethaizone:
Ambiguous field error will appear when search and join relations together.

This patch will only get table from builder class and push it before column name.

_This patch need for next pull request that I about sort in nested relations._

### DIFF
--- a/src/Engines/QueryBuilderEngine.php
+++ b/src/Engines/QueryBuilderEngine.php
@@ -384,6 +384,7 @@ class QueryBuilderEngine extends BaseEngine
          * Ambiguous field error will appear
          * when query use join table and search with keyword.
          */
+        // Remove backtick that appear from MYSQL query.
         $column = str_replace('`', '', $column);
 
         // check . in field name for protect don't add table again
@@ -400,8 +401,8 @@ class QueryBuilderEngine extends BaseEngine
             // get table from query and add it.
             $column = $q->from.'.'.$column;
         }
-        // Add ` cover table and field name.
-        $column = '`' . str_replace('.', '`.`', $column) . '`';
+        // Add wrap cover table and field name.
+        $column = $this->wrap($column);
 
         /* end fix */
         
--- a/src/Engines/QueryBuilderEngine.php
+++ b/src/Engines/QueryBuilderEngine.php
@@ -378,6 +378,33 @@ class QueryBuilderEngine extends BaseEngine
         $column = $this->castColumn($column);
         $sql    = $column . ' LIKE ?';
 
+        /**
+         * Patch for fix about ambiguous field
+         *
+         * Ambiguous field error will appear
+         * when query use join table and search with keyword.
+         */
+        $column = str_replace('`', '', $column);
+
+        // check . in field name for protect don't add table again
+        // but as far as I tested, this function has single field name only.
+        if (strpos($column, '.') === false) {
+            // Alternative method to check
+            // instanceof \Illuminate\Database\Eloquent\Builder
+            if (method_exists($query, 'getQuery')) {
+                $q = $query->getQuery();
+            } else {
+                $q = $query;
+            }
+
+            // get table from query and add it.
+            $column = $q->from.'.'.$column;
+        }
+        // Add ` cover table and field name.
+        $column = '`' . str_replace('.', '`.`', $column) . '`';
+
+        /* end fix */
+        
         if ($this->isCaseInsensitive()) {
             $sql = 'LOWER(' . $column . ') LIKE ?';
         }
--- a/src/Engines/QueryBuilderEngine.php
+++ b/src/Engines/QueryBuilderEngine.php
@@ -384,8 +384,9 @@ class QueryBuilderEngine extends BaseEngine
          * Ambiguous field error will appear
          * when query use join table and search with keyword.
          */
-        // Remove backtick that appear from MYSQL query.
-        $column = str_replace('`', '', $column);
+        // Remove delimiter of column that appear from MYSQL query.
+        $column = str_replace(['`', '"', '[', ']'], '', $column);
+
 
         // check . in field name for protect don't add table again
         // but as far as I tested, this function has single field name only.


yajra:
Wow! If this works, it will definitely fix a lot of issues in the package. Ambiguous field issue is quite common in the package and haven't been addressed well. Will try to test and review this asap. Thanks a lot!

ethaizone:
For how to make error for test.

- Create table with relation. It's not important that it must be nested relations, 1 level relation will cause bug too but in all relation must have some ambiguous field to test. For me I tested on hasMany relation, `id` will be ambiguous field.
- Try sort by field from relation. This action will change from normal query to join query.
- Search keyword in filter. This action will cause error ambiguous field because when we search, we don't point which table of each field that we will search.

yajra:
@ethaizone thanks for clarification. Can you please paste some snippets like on your `js` on how you tested this?

ethaizone:
### JS
```javascript
(function(window, $) {
    window.LaravelDataTables = window.LaravelDataTables || {};
    window.LaravelDataTables["task_table"] = $("#task_table").DataTable({
        "serverSide": true,
        "processing": true,
        "ajax": {
            "url": "http:\/\/local.dev\/tasks\/data"
        },
        "columns": [{
            "data": "id",
            "title": "task_id",
            "orderable": true,
            "searchable": true
        }, {
            "data": "block.id",
            "title": "block_id",
            "orderable": true,
            "searchable": true
        }, {
            "data": "updated_at",
            "title": "updated_at",
            "orderable": true,
            "searchable": true
        }],
        "bAutoWidth": false
    });
})(window, jQuery);
```

### Table structure
(I will skip not important fields.)
**tasks**
- id
- block_id
**blocks**
- id

### Model
```php
namespace App;

use Illuminate\Database\Eloquent\Model;

class Task extends Model
{
    public function block()
    {
        return $this->belongsTo(Block::class);
    }
}

class Block extends Model {}
```

### Controller
```php
use Yajra\Datatables\Html\Builder;

    public function getIndex(Builder $datatableBuilder)
    {
        $datatables = $datatableBuilder
            ->setTableAttribute('id', 'task_table')
            ->addColumn(['data' => 'id', 'title' => __a('task_id')])
            ->addColumn(['data' => 'block.id','title' => __a('block_id')])
            ->addColumn(['data' => 'updated_at', 'title' => __a('updated_at')])
            ->ajax(['url' => route('tasks.index-data.get')]);
        return view('pages.tasks.index', compact('datatables'));
    }

    public function getIndexData(Request $request)
    {
        $tasks = app(Task::class)->query()
            ->with(['block']);

        $datatables = Datatables::of($tasks)
            ->editColumn('id', function($task) {
                return $task->short_id;
            })
            ->editColumn('block.id', function($task) {
                return $task->block->short_id;
            });
        return $datatables->make(true);
    }
```

### Query string from ajax
```
draw:3
columns[0][data]:id
columns[0][name]:
columns[0][searchable]:true
columns[0][orderable]:true
columns[0][search][value]:
columns[0][search][regex]:false
columns[1][data]:block.id
columns[1][name]:
columns[1][searchable]:true
columns[1][orderable]:true
columns[1][search][value]:
columns[1][search][regex]:false
columns[2][data]:updated_at
columns[2][name]:
columns[2][searchable]:true
columns[2][orderable]:true
columns[2][search][value]:
columns[2][search][regex]:false
order[0][column]:1
order[0][dir]:asc
start:0
length:10
search[value]:60001
search[regex]:false
_:1479992152553
```

### Error
```
QueryException in Connection.php line 729:
SQLSTATE[23000]: Integrity constraint violation: 1052 Column 'id' in where clause is ambiguous (SQL: select * from `tasks` left join `blocks` on `tasks`.`block_id` = `blocks`.`id` where (LOWER(`id`) LIKE %60001% or (select count(1) from `blocks` where `tasks`.`block_id` = `blocks`.`id` and LOWER(`id`) LIKE %60001%) >= 1 or LOWER(`updated_at`) LIKE %60001%) order by `blocks`.`id` asc limit 10 offset 0)
```

I just in happy mood so I will provide all data. Hahaha.

This patch can fix about search in sorted relations. This patch will replace ``LOWER(`id`)`` to be ``LOWER(`tasks`.`id`)``. Error occur because we have `left join` via ``order by `blocks`.`id` asc`` so this patch will help this case.

src/Engines/QueryBuilderEngine.php
```diff
@@ -378,6 +378,33 @@ protected function compileQuerySearch($query, $column, $keyword, $relation = 'or
...
+                $q = $query->getQuery();
+            } else {
+                $q = $query;
+            }
+
+            // get table from query and add it.
+            $column = $q->from.'.'.$column;
+        }
+        // Add ` cover table and field name.
+        $column = '`' . str_replace('.', '`.`', $column) . '`';
```
ethaizone: I see. I will look at it. I use MYSQL only so I don't test another DB.

src/Engines/QueryBuilderEngine.php
```diff
@@ -378,6 +378,34 @@ protected function compileQuerySearch($query, $column, $keyword, $relation = 'or
...
         $sql    = $column . ' LIKE ?';
 
+        /**
+         * Patch for fix about ambiguous field
+         *
+         * Ambiguous field error will appear
+         * when query use join table and search with keyword.
+         */
+        // Remove backtick that appear from MYSQL query.
+        $column = str_replace('`', '', $column);
```
yajra: Is this part necessary? Other DB uses `"` double quote to escape the columns. I think MSSQL uses brackets like `[column]`?

ethaizone: I just test on MYSQL only so I found ` only.

I don't know for other DB but I'm sure that `$column` has delimiter for column. I need to remove it before add table name.

After this I need type in code because github markdown and backtick conflict.

```
Example $column is `field`. If I don't remove backtick and add table name.
$column will be table_name.`field` and it will go in wrap method.
Result will look like `table_name`.``field``.

It will error.
```

I don't know which solution to remove delimiter from column name. Do you have any suggestion?

yajra: Not sure but maybe we could try replacing those 4 symbols with empty string? Will check this further on my oracle project when I had the chance. Thanks!

```php
$column = str_replace(['`[]"'], '', $column);
```

ethaizone: I will update code to remove this 4 symbols.

yajra:
🚀  Released on v6.22.2. Thanks!

PR #878, opened 2016-12-06
Custom soft delete column support

Huracan88:
Get the right column for soft delete via Laravel model method.

### DIFF
--- a/src/Engines/QueryBuilderEngine.php
+++ b/src/Engines/QueryBuilderEngine.php
@@ -103,7 +103,7 @@ class QueryBuilderEngine extends BaseEngine
 
         // check for select soft deleted records
         if (! $this->withTrashed && $this->modelUseSoftDeletes()) {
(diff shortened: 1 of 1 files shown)


src/Engines/QueryBuilderEngine.php
```diff
@@ -103,7 +103,7 @@ public function count()
 
         // check for select soft deleted records
         if (! $this->withTrashed && $this->modelUseSoftDeletes()) {
-            $myQuery->whereNull($myQuery->getModel()->getTable() . '.deleted_at');
+            $myQuery->whereNull($myQuery->getModel()->getQualifiedDeletedAtColumn());
```
yajra: We might need to still include the table name to avoid ambiguous column issues:

```php
$myQuery->whereNull( $myQuery->getModel()->getTable() . '.' . $myQuery->getModel()->getQualifiedDeletedAtColumn());

```

Thanks!

PR #934, opened 2017-01-20
Fix regex query for pgsql.

alfa6661:
Getting error when using pgsql database and activate the regex filtering.

```
SQLSTATE[42601]: Syntax error: 7 ERROR:  syntax error at or near "REGEXP"
LINE 1: ...s "row_count" from "categories" where LOWER(name) REGEXP $1)...
```

on postgreSQL there is no REGEXP command. We can use [POSIX regular expression](https://www.postgresql.org/docs/9.3/static/functions-matching.html#FUNCTIONS-POSIX-TABLE)


yajra:
Thanks for the PR. Will try to test and review this ASAP. 👍

yajra:
Released on v6.24.1 🚀  Thanks a lot!

PR #962, opened 2017-02-02
[7.0] Add support for 'where' in the relationship

snagytx:
Pull the existing bindings from the builder and prepare the first element based on our configuration. Doing so we also pull any extra bindings that the builder has.

Dow now this was tested ONLY with relationships similar to:

```php
        return $this->hasOne('App\FieldOption', 'sort_order', 'service')
            ->where('group_id', '=', '1');
```

### DIFF
--- a/src/Engines/QueryBuilderEngine.php
+++ b/src/Engines/QueryBuilderEngine.php
@@ -341,7 +341,7 @@ class QueryBuilderEngine extends BaseEngine
                 $relationMethod = "whereRaw";
             }
 
-            //owerwrite the first element of the binding - we expect the keyword to be located there
+            //overwrite the first element of the binding - we expect the keyword to be located there
             $bindings[0]  = $this->prepareKeyword($keyword);
             
             if ($relationType instanceof MorphToMany) {
--- a/src/Engines/QueryBuilderEngine.php
+++ b/src/Engines/QueryBuilderEngine.php
@@ -329,6 +329,7 @@ class QueryBuilderEngine extends BaseEngine
             $builder      = $chunk['builder'];
             $relationType = $chunk['relationType'];
             $query        = $chunk['query'];
+            $bindings     = $builder->getBindings();
             $builder      = "({$builder->toSql()}) >= 1";
 
             // Check if it last relation we will use orWhereRaw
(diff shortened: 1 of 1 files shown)


yajra:
@snagytx thanks, I was able to look at this yesterday and have found a different solution. Will send my own PR later but will also test this one. Thanks!

yajra:
@snagytx can you please verify too if my PR #968 works on your side? Thanks!

src/Engines/QueryBuilderEngine.php
```diff
@@ -340,10 +341,13 @@ protected function compileRelationSearch($query, $relation, $column, $keyword)
                 $relationMethod = "whereRaw";
             }
 
+            //overwrite the first element of the binding - we expect the keyword to be located there
+            $bindings[0]  = $this->prepareKeyword($keyword);
```
yajra: I find this to be a bit of a hack. What if we have some where statement from the base query?

snagytx:
@yajra I guess you could say that it's a bit hackish, I thought that the like is being build somewhere else but looking at `compileQuerySearch` I see that `prepareKeyword` is already being called on the keyword we don't need to re-add it - maybe we can skip this hackish line.

I didn't do this since I don't have this type of relationships to test but for MorphToMany it would be required to merge the `$relationType->getMorphClass()` and `$bindings` arrays together in the right order - but this would need some testing.

src/Engines/QueryBuilderEngine.php
snagytx: @yajra I guess you could say that it's a bit hackish, I thought that the 'LIKE' is being build somewhere else but looking at `compileQuerySearch` I see that `prepareKeyword` is already being called on the keyword we don't need to re-add it - maybe we can skip this hackish line.

I didn't do this since I don't have this type of relationships to test but for `MorphToMany` it would be required to merge the `$relationType->getMorphClass()` and `$bindings` arrays together in the right order - but this would need some testing.

snagytx:
The proposed changes were testing using:

```
 $query = WelcomeCard::with(['service_rel', ...])
            ->where('id', '<', 100)
            ->select(['welcome_cards.*']);
```

```
    public function service_rel(){
        return $this->hasOne('App\FieldOption', 'sort_order', 'service')
            ->where('group_id', '=', 1);
    }
```

src/Engines/QueryBuilderEngine.php
yajra: Just tested the PR and it seems like this line is not need and can be removed. Will merge and do the necessary changes. Thanks for the hint!

PR #850, opened 2016-11-22
[6.0] Add support ordering when search in nested relations.

ethaizone:
I did it!!

PR #965, opened 2017-02-02
[7.0] Add support ordering when search in nested relations

AdrienPoupa:
Same commit as #850, updated for v7

I don't know if this is related, but I had to change

$columns = $this->request->columns();

back to

$columns = (array) $this->request->input('columns');

for some datatables (method not supported)

### DIFF
--- a/src/Engines/QueryBuilderEngine.php
+++ b/src/Engines/QueryBuilderEngine.php
@@ -517,42 +517,52 @@ class QueryBuilderEngine extends BaseEngine
      */
     protected function joinEagerLoadedColumn($relation, $relationColumn)
     {
-        $model = $this->query->getRelation($relation);
-        switch (true) {
-            case $model instanceof BelongsToMany:
-                $pivot   = $model->getTable();
-                $pivotPK = $model->getExistenceCompareKey();
-                $pivotFK = $model->getQualifiedParentKeyName();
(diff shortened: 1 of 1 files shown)


yajra:
> $columns = $this->request->columns();

This method was added only on v7.

Anyways, thanks for the PR. Will try to test and review this as soon as I can. 

Note: One issue I found on the previous PR was when the relation uses self join. Maybe you can try it too when you have time. Thanks!

AdrienPoupa:
`This method was added only on v7.`
I know, but I faced this bug using v7, that is what I found weird. Anyway, I will test it :)

yajra:
@AdrienPoupa FYI, this is my [test repo](https://github.com/yajra/dt54) for Laravel 5.4 with v7.0. Maybe you can also use this to possibly test this feature and maybe also contribute it to docs.. Thanks a lot! ^_^

AdrienPoupa:
Editable docs? Wow that's a good idea, I always found a bit frustrating that current document is not

yajra:
The main docs located [here](https://github.com/yajra/laravel-datatables-docs) too in-case you need it.

yajra:
Released on v7.3.0 🚀 

Will address the self join issue on a future releases. Thanks!

PR #992, opened 2017-02-09
[7.0] Add support for array data source.

yajra:
Idea came from this [Laracast Forum](https://laracasts.com/discuss/channels/general-discussion/datatables-from-a-raw-query) question.

Examples:

```php
$source = [
   [1, 2, 3],
   [4, 5, 6],
];

return Datatables::of($source)->make();
```

```php
$source = DB::select('select * from users');

return Datatables::of($source)->make(true);
```

### DIFF
--- a/src/Datatables.php
+++ b/src/Datatables.php
@@ -2,7 +2,7 @@
 
 namespace Yajra\Datatables;
 
-use Yajra\Datatables\Html\Builder;
+use Illuminate\Support\Collection;
 
 /**
  * Class Datatables.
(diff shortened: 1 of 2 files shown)


jim1506:
This is a really useful addition. Thanks so much for your work.

kamleshwebtech:
Below both solutions are working:

/* By DB Table and querybuilder */
$query = DB::table('clients')->where('email','like','%gmail.com');
return app('datatables')->queryBuilder($query)->toJson();

/* Core query */
$source = DB::select('select * from clients');
return Datatables::of($source)->make(true);

mateusfig:
This causes overload working. From a big database table will yield slowness. Could not use this solution of course.

PR #995, opened 2017-02-10
Adding Multiple Table Rendering Feature

OzanKurt:
Hi, I added support for rendering 2 tables in a single page.

The idea was pretty simple. 

I had 2 tables in `/users`, `ActiveUsersDataTable` and `InactiveUsersDataTable`.

- I added and `id` field to use in html part of the datatable so that the ids won't conflict.
- I added `?tableId=TableID` to each request so that we can define which table we're querying with the ajax request to the backend.
- I edited the `server-side.buttons.js` to allow urls with a query parameter.

### DIFF
--- a/src/Datatables.php
+++ b/src/Datatables.php
@@ -105,6 +105,48 @@ class Datatables
     }
 
     /**
+     * Process dataTables needed render output.
+     *
+     * @param array $tables
+     * @param string $view
+     * @param array $data
+     * @param array $mergeData
+     * @return \Illuminate\Http\JsonResponse|\Illuminate\View\View
+     */
+    public function renderMultiple(array $tables, $view, $data = [], $mergeData = [])
+    {
+        $tableId = $this->request->input('tableId');
+
+        if ($this->request->ajax() && $this->request->wantsJson()) {
+            return $this->getTableFromId($tables, $tableId)->ajax();
+        }
+
+        $action = $this->request->get('action');
+
+        if (in_array($action, ['print', 'csv', 'excel', 'pdf'])) {
+            if ($action == 'print') {
+                return $this->getTableFromId($tables, $tableId)->printPreview();
+            }
+
+            return call_user_func_array([$this->getTableFromId($tables, $tableId), $action], []);
+        }
+
+        $tables = array_map(function($table) {;
+            return $table->html();
+        }, $tables);
+
+        return view($view, $data, $mergeData)->with($tables);
+    }
+
+    private function getTableFromId($tables, $tableId)
+    {
+        if (!array_key_exists($tableId, $tables)) {
+            throw new \Exception("Table {$tableId} is not defined in multiple renderer.");
+        }
+        return $tables[$tableId];
+    }
+
+    /**
      * Get html builder instance.
      *
      * @return \Yajra\Datatables\Html\Builder
--- a/src/Datatables.php
+++ b/src/Datatables.php
@@ -131,13 +131,21 @@ class Datatables
             return call_user_func_array([$this->getTableFromId($tables, $tableId), $action], []);
         }
 
-        $tables = array_map(function($table) {;
+        $tables = array_map(function($table) {
+            if (!$table->hasCustomId()) {
+                $table->setId($this->createCustomIdFor($table));
+            }
             return $table->html();
         }, $tables);
 
         return view($view, $data, $mergeData)->with($tables);
     }
 
+    private function createCustomIdFor($table)
+    {
+        return (new \ReflectionClass($table))->getShortName();
+    }
+    
     private function getTableFromId($tables, $tableId)
     {
         if (!array_key_exists($tableId, $tables)) {
--- a/src/Datatables.php
+++ b/src/Datatables.php
@@ -143,7 +143,7 @@ class Datatables
 
     private function createCustomIdFor($table)
     {
-        return (new \ReflectionClass($table))->getShortName();
+        return camel_case((new \ReflectionClass($table))->getShortName());
     }
     
     private function getTableFromId($tables, $tableId)


ManuDoni:
Thumbs up!
I think this will be very useful.
I will use this package only if this will be merged because in my project I have a table inside every tabs

OzanKurt:
:)

Ozan Kurt

yajra:
@OzanKurt I think this PR should go to buttons plugin repo? Can you please provide some snippets on how you use this? I think this is implemented along with `DataTable` service class? thanks!

OzanKurt:
Yes, this changes require some changes in both packages.

Here is a usage example thou:

DoctorInvitationsController.php
```php
    public function index(DoctorInvitationsDataTable $doctorInvitationsDataTable, InvitableDoctorsDataTable $invitableDoctorsDataTable)
    {
        return Datatables::renderMultiple([
            'invitableDoctorsDataTable' => $invitableDoctorsDataTable,
            'doctorInvitationsDataTable' => $doctorInvitationsDataTable,
        ], 'clinicforce.doctor-invitations.index');
    }
```

index.blade.php
```blade
@section('content')
    {!! $invitableDoctorsDataTable->table() !!}
    {!! $doctorInvitationsDataTable->table() !!}
@stop

@section('scripts')
    {!! $invitableDoctorsDataTable->scripts() !!}
    {!! $doctorInvitationsDataTable->scripts() !!}
@stop
```

It is being so complicated with my changes being sent as pull requests on each repo individially. I can send you the edited `vendor/` directory, what do you think?

yajra:
@OzanKurt thanks for the snippets, make more sense now. However, I think this should indeed be on the buttons repo package. Maybe it would be better if we can create something like `Manager` class that handles this edge cases. This way, we will not clog up our main repo for some additional stuffs which was the reason why we extracted functionalities to separate packages.

```php
DataTableManager::render([
            'invitableDoctorsDataTable' => $invitableDoctorsDataTable,
            'doctorInvitationsDataTable' => $doctorInvitationsDataTable,
        ], 'clinicforce.doctor-invitations.index');
```

Also, it seems like this function does not really need the functions inside the `Datatables` class? Thus, this can be promoted to it's own class handler? What do you think?

BTW, this addition would be really great. Looking forward to this feature. Thanks a lot! 👍

OzanKurt:
I am currently using my version in a project. Manager seems a good idea, but tbh I don't know how I might create a manager.

Ozan Kurt

Omranic:
@yajra any chance this could be merged anytime soon?

yajra:
@Omranic Sorry but I'm a bit occupied and cannot test this PR. If you can, please provide some feedback on how your testing goes? I would gladly merge this if it works for you. Thanks!

lk77:
Hello,

this belongs to datatable-buttons package i think,
We could create a new service like DataTableManager, which will handle this edge case.
with a method like that :
```
public function renders($view, array $parameters, array $dataTables)
    {
       if ($this->request->ajax() && $this->request->has('tableId')) {
            return $dataTables[$this->request->input('tableId')]->ajax();
        } else {
            return view($view, array_merge($parameters, $this->htmlDataTable($dataTables)));
        }
}
```
and like that
```
private function htmlDataTable(array $dataTables)
    {
        foreach ($dataTables as $key => $dataTable) {
            $dataTables[$key] = $dataTable->html();// return  the builder from datatable-html
        }

        return $dataTables;
    }
```

chrisidakwo:
What's the update on this? Or has it been implemented outside of this PR?

yajra:
@chrisidakwo unfortunately I wasn't able to provide much time to review this PR. 

On the other hand, I implemented cases like this by separating each DataTable to it's own controller. Using the example above:

```php
class UsersController {
  public function index(InactiveUsersDataTable $inactiveDataTable, ActiveUsersDataTable $dataTable) {
    $active = $activeDataTable->html();
    $inactive = $inactiveDataTable->html();

    return view('users.index', compact('active', 'inactive'));
  }
}

class InactiveUsersController {
  public function index(InactiveUsersDataTable $dataTable) {
    return $dataTable->render('users.inactive.index');
  }
}

class ActiveUsersController {
  public function index(ActiveUsersDataTable $dataTable) {
    return $dataTable->render('users.active.index');
  }
}

// users.index blade
@section('content')
    {!! $active->table() !!}
    {!! $inactive->table() !!}
@stop

@section('scripts')
    {!! $active->scripts() !!}
    {!! $inactive->scripts() !!}
@stop
```

For the DataTable, you just need to specify the table id and ajax to the specific route.

```php
->setTableId('inactive-users')
->ajax(route('inactive-users.index'))
...
->setTableId('active-users')
->ajax(route('active-users.index'))
```

ATM, I prefer this approach because I can easily promote `InactiveUsers` to it's own page if needed.

yajra:
In addition, I am also tinkering about separating the `html()` part of the service to its own class. This way, the builder can be used directly without having to inject the full DataTable class. Already have a WIP on my local machine and still polishing it.

```php
class ActiveUsersDataTable extends DataTable {
  public function htmlBuilder() : Builder {
    return new ActiveUsersDataTableHtml;
  }
}

class InactiveUsersDataTable extends DataTable {
  public function htmlBuilder() : Builder {
    return new InactiveUsersDataTableHtml;
  }
}

// Controller
class UsersController {
  public function index(ActiveUsersDataTableHtml $active, InactiveUsersDataTableHtml $inactive) {
    return view('users.index', compact('active', 'inactive'));
  }
}
```

yajra:
@OzanKurt @lk77 @chrisidakwo @Omranic, @ManuDoni sent a PR https://github.com/yajra/laravel-datatables-buttons/pull/99. 

Requesting a review if possible. My proposal would break the current behaviour for the service class but is still backward compatible. 

In relation to this PR, the proposed solution would be used like:

```php
class UsersController {
  public function index() {
    $active = ActiveUsersDataTableHtml::make();
    $inactive = InactiveUsersDataTableHtml::make();

    return view('users.index', compact('active', 'inactive'));
  }
}
```

OzanKurt:
@yajra Hey,

Did you also think about constructor parameters when using these tables?

I am using the datatables in routes like this very often: `/users/{user}/documents`

I customised the constructor to adapt the URL according to the parameter I pass in. So then I am simply able to use `$this->user` and therefore customize everything according to it.

```php
/**
 * UsersBaseDataTable
 * @package App\DataTables\Users
 */
class UsersBaseDataTable extends BaseDataTable
{
    /**
     * User instance to be used for table data.
     *
     * @var \App\Models\User|null
     */
    protected $user;

    /**
     * Construct UsersBaseDataTable
     *
     * @param \App\Models\User $user
     */
    public function __construct(User $user)
    {
        parent::__construct();

        $this->user = $user;

        $this->replaceUrlParameter('user', $user->id ?? 0);
    }
}

/**
 * DocumentsDataTable
 * @package App\DataTables\Users
 */
class DocumentsDataTable extends UsersBaseDataTable
{
    /**
     * Table id for html builder.
     *
     * @var string
     */
    protected $id = 'usersDocumentsDataTable';

    /**
     * Ajax url, to retrieve table content.
     *
     * @var string
     */
    protected $url = '/datatables/users/{user}/documents';

    public $deferLoading = true;

    /**
     * Construct DocumentsDataTable
     *
     * @param User $user
     */
    public function __construct(User $user)
    {
        parent::__construct($user);
    }
}
```

What do you think?

OzanKurt:
Also maybe we also need to think of a solution to keep both HtmlBuilder and DataTable class, in the same class as it is right now. I know a lot of developers who get overwhelmed by extra classes.

yajra:
@OzanKurt I think I haven't use dataTables that way. When I need some dependencies from controller for instance, I always go to custom setter approach. Something like below:

```php
Route::resource('order.details', 'OrderDetailsController');

class OrderDetailsController {
    public function index(OrderDetailsDataTable $dataTable, Order $order) {
        return $dataTable->setOrder($order)->render('view');
    }
}

class OrderDetailsDataTable extends DataTable {
    protected $order;

    public function setOrder(Order $order) {
        $this->order = $order;
        return $this;
    }
}
```

> Also maybe we also need to think of a solution to keep both HtmlBuilder and DataTable class, in the same class as it is right now. I know a lot of developers who get overwhelmed by extra classes.

Agree a bit on this. After using the HtmlBuilder for quite some time now, I have been reverting back to old ways where all is inside a single class specially for simple tables. I think the new builder class should only be used on some edge cases where the html needs to be reused. Anyways, it's still optional to use that approach.

OzanKurt:
Yes exactly, a base query with a lot of differed column setups might need a separate HtmlBuilder, which makes A LOT of sense. I am going to update to latest version of DataTables once I have some free time.

```php
class OrderDetailsDataTable extends DataTable {
    protected $order;

    public function setOrder(Order $order) {
        $this->order = $order;
        return $this;
    }
}
```

This code you send would be a really simple way. I will play with this a bit since concstructors and dependenct injection can sometimes be pain in Laravel. I will let you know.

yajra:
Sorry for not being able to check this and its now outdated. Let's try to implement again on v10 when get the chance. Thanks!

PR #998, opened 2017-02-13
[7.0] Added search multi terms really smart.

zeno-mioso:
Hi, 

i tried to implement a solution that care about multi terms search.

Please, take a look. Thanks for your work, I really appreciate it @ work!

### DIFF
--- a/src/Engines/QueryBuilderEngine.php
+++ b/src/Engines/QueryBuilderEngine.php
@@ -354,15 +354,28 @@ class QueryBuilderEngine extends BaseEngine
      */
     protected function compileQuerySearch($query, $column, $keyword, $relation = 'or')
     {
-        $column = $this->addTablePrefix($query, $column);
-        $column = $this->castColumn($column);
-        $sql    = $column . ' LIKE ?';
+        $keyword = trim($keyword);
+        if (false===strpos($keyword,' ')) {
+            $column = $this->addTablePrefix($query, $column);
+            $column = $this->castColumn($column);
+            $sql    = !$this->isCaseInsensitive() ? $column . ' LIKE ? ':'LOWER(' . $column . ') LIKE ? ';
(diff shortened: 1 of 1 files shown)


yajra:
@zeno-mioso thanks for the great idea. 💯 

On the other hand, I think this will fail on some database? I think `concat_ws` is only applicable on MySQL? Please consider the PR to work on other database too that Laravel supports.

zeno-mioso:
I tried to use `concat` instead `concat_ws` but travis fired me! :(
Any suggest to complete my commit?

Thanks

yajra:
Sorry for late response. Will try to test this as soon as I can. 

Question: Is concat available on most databases? Can't find the right term to search on google :(

We need to be careful since this might cause a breaking change. Thanks!

yajra:
@zeno-mioso as per travis, it seems like SQLITE does not support concat.

```
Caused by
PDOException: SQLSTATE[HY000]: General error: 1 no such function: concat
```

apreiml:
If I understood your changes correctly, you concat all columns together for each column and then query this concatenation with each of the keywords. But that means you concat the same strings and do the same comparison for each of the columns.

if you have for example 3 columns (a, b, c) and two keywords (x, y) you'll generate a query like:
```
where
  concat(a, ' ', b, ' ', c) like x and concat(a, ' ', b, ' ', c) like y and
  concat(a, ' ', b, ' ', c) like x and concat(a, ' ', b, ' ', c) like y and
  concat(a, ' ', b, ' ', c) like x and concat(a, ' ', b, ' ', c) like y 
```
which is imho too much concats and also the code is very hard to understand, for what it does.

apreiml:
I'd like to suggest a simpler solution for this problem: #1113

yajra:
Implemented on #1113. Thanks!

PR #1026, opened 2017-02-22
bump to 0.8 dompdf for compatability with barryvdh/dompdf

MasterBratac:
Bumped dompdf version to 0.8 for compatability with barryvdh/laravel-dompdf

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,7 +19,7 @@
         "league/fractal": "~0.14",
         "laravelcollective/html": "5.0.*|5.1.*|5.2.*|5.3.*",
         "maatwebsite/excel": "^2.0",
-        "dompdf/dompdf": "^0.7"
+        "dompdf/dompdf": "^0.8"
     },
     "require-dev": {
         "mockery/mockery": "~0.9",


mikkokut:
+1

yajra:
Thanks @MasterBratac, does exporting of pdf still works when updating to `0.8`? Anyways, will try to do some testing within the week. Thanks!

yajra:
I made some tests and it seems like exporting of PDF using DOMPDF are not working even on `v0.7.0`. I think we should completely remove `dompdf` dependency and let the developers use their desired version?

This would be a breaking change but I think is necessary? What do you think?

Referencing [Laravel Excel](http://www.maatwebsite.nl/laravel-excel/docs/export), the recommended version is `~0.6.2`.

yajra:
Made further research and it seems like `~0.6.2` is not working on PHP7.1 but compatibility was fixed on `0.8.0`. However `PHPExcel` does not support version `> 0.6.2`. Based on this findings, I guess it is better to remove the dependency and just let the developers decide and implement their ways to exporting to PDF. Thanks!

yajra:
Released on v6.26.0 but dependency on `dompdf` was removed. Thanks!

Issue #1070, opened 2017-03-17
Display url or balise A as a link in Excel export


julianfox:
### Summary of problem or feature request
I just managed to display all my html balise A as a proper Excel hyperlink during export.
All is working perfectly but I don't know if my approach is correct:

### Code snippet
I extend decodeContent of Yajra\Datatables\Transformers\DataTransformer
for transform A balise to url.
```php
/**
     * Decode content to a readable text value.
     *
     * @param string $data
     * @return string
     */
    protected function decodeContent($data)
    {
        preg_match_all('/<a[^>]+href=([\'"])(.+?)\1[^>]*>/i', $data, $result);
        if ($result[2] && $result[2][0]) {
            return $result[2][0];
        }

        $decoded = html_entity_decode(strip_tags($data), ENT_QUOTES, 'UTF-8');
        return str_replace("\xc2\xa0", ' ', $decoded);
    }
```

And I extend buildExcelFile of Yajra\Datatables\Services\Datatable
for transform the url into proper Excel hyperlink
```php
/**
     * Build excel file and prepare for export.
     *
     * @return mixed
     */
    protected function buildExcelFile()
    {
        return app('excel')->create($this->getFilename(), function (LaravelExcelWriter $excel) {
            $excel->sheet('exported-data', function (LaravelExcelWorksheet $sheet) {
                $sheet->fromArray($this->getDataForExport());
                $highestColumn = $sheet->getHighestColumn();

                for($row = 1; $row <= sizeof($sheet->data); $row++) {
                    foreach (range('A', $highestColumn) as $column) {
                        $cell = $sheet->getCell($column.$row);
                        $cellValue = $cell->getValue();

                        if(filter_var($cellValue, FILTER_VALIDATE_URL)) {
                            $cell->setValue($cellValue)
                                 ->getHyperlink()
                                 ->setUrl($cellValue);
                        }
                    }
                }
            });
        });
    }
```

Is my approach correct ?
If yes is it useful to do a PR ? I would be glad to help :-)

yajra:
@julianfox the code looks fine with me and I think would be a great addition to be able to export links by default. Can you please submit a PR and will test/review it when I had the chance. Thanks a lot!

julianfox:
Done #1074 :-)

PR #1074, opened 2017-03-18
[6.0] Display url and A balise as hyperlink in excel export

julianfox:
Following the feature request: #1070

### DIFF
--- a/src/Services/DataTable.php
+++ b/src/Services/DataTable.php
@@ -12,6 +12,8 @@ use Yajra\Datatables\Contracts\DataTableScopeContract;
 use Yajra\Datatables\Datatables;
 use Yajra\Datatables\Transformers\DataTransformer;
 
+use SimpleXMLElement;
+
 /**
  * Class DataTable.
  *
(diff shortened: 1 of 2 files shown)


yajra:
Thanks, will review this as soon as I can. 👍

yajra:
@julianfox I made some test and I think there is some minor issue on exporting the link to excel.

As per testing, I converted the id to link. The link was exported successfully but the value of the id was also changed using the link path.

Example output:

```
| http://datatables.dev | Elmore Howe | kgreenfelder@bode.com |
| http://datatables.dev | Ashley Johnston	| mharber@yahoo.com |
```

Expected:

```
| 1 > with link to http://datatables.dev | Elmore Howe | kgreenfelder@bode.com |
| 2 > with link to http://datatables.dev | Ashley Johnston	| mharber@yahoo.com |
```

julianfox:
Yes, it's a normal behaviour, because in my case, we use font awesome icon as value in the datatables. But yes in most of the cases that make sense to keep the original value I'll fix it as soon as I can 👍

julianfox:
I just did an update.

I use the balise a text value as excel file cell value now, only if the balise a text value is a string.
If the balise a text value is html (e.g. button), I use the href value as cell value.

yajra:
Works like a charm. Can you please also send the same PR to https://github.com/yajra/laravel-datatables-buttons if possible for same support on v7? 

Released on v6.28.0 🚀  Thanks!

julianfox:
Ok I will check it as soon as I can

Issue #1109, opened 2017-04-18
Smart Search Documentation Issue


apreiml:
### Summary of problem or feature request

The Documentation says that smart search enables the Datatables smart search feature, but in this project it is understood differently. The Datatables documentation says that smart search splits the string into individual words and searches for each of them. This project on the other hand just wraps the string with '%' chars.

See BaseEngine::setupKeywords

What would it take to implement this feature properly? Can I somehow help?

This may also be the problem for #881 .

apreiml:
Seems like https://github.com/yajra/laravel-datatables/pull/998 is trying to achieve this.

yajra:
@apreiml yes, the smart search implemented atm are just a wildcard database search. If you can test PR #998 on popular databases then if would really be a big help. We should try to support as much database as possible. Thanks!

yajra:
Fixed on v7.4.0. Thanks!

PR #1113, opened 2017-04-19
[7.0] Implement multi-term smart search in QueryBuilderEngine.

apreiml:
This pr tries to solve the smart search feature #1109 . The idea for the requested change is from my colleague.

if this solution is preferable I'd also try to implement smart search for the collection engine.

### DIFF
--- a/src/Engines/QueryBuilderEngine.php
+++ b/src/Engines/QueryBuilderEngine.php
@@ -126,9 +126,17 @@ class QueryBuilderEngine extends BaseEngine
      */
     public function filtering()
     {
+         $keywords = array_filter(explode(' ', $this->request->keyword()));
+
+         foreach ($keywords as $keyword) {
+            $this->globalSearch($keyword);
+         }
+    }
+
+    private function globalSearch($keyword)
+    {
         $this->query->where(
-            function ($query) {
-                $globalKeyword = $this->request->keyword();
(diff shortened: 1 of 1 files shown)


yajra:
Wow, nice PR and looks good. 💯 

Please go ahead and do the same for collection engine :). Will test and review this as soon as I can. Thanks!

apreiml:
Smart search is now also implemented for the CollectionEngine. It is now also disabled if smart is set to false.

One thing that still bugs me though is, that QueryBuilder doesn't wrap the string with '%', smart search is disabled. The CollectionEngine on the other hand uses string contains, which has similar functionality as wrapping the string with '%' and using `like` in a query. So it behaves like the old smart search, even if smart search is disabled.

Do you want to really omit the '%', if smart search is disabled? Or does the CollectionEngine need to match the collumns exactly, if smart search is disabled?

yajra:
@apreiml thanks for the updates and can you please create a separate PR for collection engine for easier testing and reviewing.

With regard to collection, I honestly rarely use collection and haven't give it so much attention. Will give it a further digging once I had the chance to review your PR. Thanks a lot!

yajra:
Released on v7.4.0 🚀 along with collection engine. This is indeed a great addition, thanks a lot! ❤️

PR #1131, opened 2017-05-05
[7.0] Add server-side error handler.

yajra:
* When app is in production, it is better to display a custom error message to the user.

* Implements server-side error handling as defined on https://datatables.net/manual/server-side.

> Optional: If an error occurs during the running of the server-side processing script, you can inform the user of this error by passing back the error message to be displayed using this parameter. Do not include if there is no error.

* Add dataTables exception class.


yajra:
Docs located at https://yajrabox.com/docs/laravel-datatables/7.0/error-handler.

PR #1158, opened 2017-05-25
bug fixed: addColumn()

liub1993:
The function **addColumn()** would throw an exception if the newly added column name can not be found in MySQL table and the query still contains the column added by **addColumn( )**. 

So I just add the new column in the blacklist, and it works.

### DIFF
--- a/src/Engines/BaseEngine.php
+++ b/src/Engines/BaseEngine.php
@@ -65,14 +65,14 @@ abstract class BaseEngine implements DataTableEngineContract
      * @var array
      */
     protected $columnDef = [
-        'index'     => false,
-        'append'    => [],
-        'edit'      => [],
-        'excess'    => ['rn', 'row_num'],
-        'filter'    => [],
-        'order'     => [],
-        'escape'    => '*',
-        'raw'       => ['action'],
+        'index' => false,
+        'append' => [],
+        'edit' => [],
+        'excess' => ['rn', 'row_num'],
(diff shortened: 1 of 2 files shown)


liub1993:
Yes, it only fails when we use the QueryBuilderEngine.

yajra:
Maybe we can put `addColumn` override inside `QueryBuilderEngine`. `addToBlacklist` can stay on `BaseEngine`.

Anyways, will review/test this asap. Thanks!

yajra:
BTW, please format the code to PSR-2.

liub1993:
Sorry for my mistake, and thanks for your great work!

liub1993:
Hi, @yajra , I have: 
1. Renamed the  `addToBlacklist`  to  `pushToBlacklist`  and removed the redundant code in the method;
2. Rewrote the `addColumn` in QueryBuilderEngine.php.

I don't know if there are any better solutions, forgive me for my pool programming ...

yajra:
@liub1993 thanks for the updates. Will do some testing but we need to be careful on this. I just realized that this may cause some **breaking change** on some users like those who are using `filterColumn` api to override the default search for added columns.

P.S. 
Can you please commit only the methods (`pushToBlacklist` and `addColumn`) affected by the updates?

yajra:
Nvm my previous comment. Merging and will do the necessary adjustments to fix the potential breaking change on #1169. Thanks!

PR #1190, opened 2017-06-06
Support Auto-Discovery for Laravel 5.5

mikimaine:
This will make the package Support Auto-Discovery in Laravel 5.5


yajra:
Thanks but I think PR #1208 implementation is more complete. Lets proceed from there!

PR #1218, opened 2017-06-20
[5.5] Laravel 5.5 Packages

leyluj:
Changes the composer file to pull the 5.5.* version of L5

Basically, when I import this package on a L5.5 it breaks so i have changes the required version for L5.5.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -15,11 +15,11 @@
   ],
   "require": {
     "php": ">=5.6.4",
-    "illuminate/database": "5.4.*",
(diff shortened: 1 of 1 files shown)


yajra:
Thanks but this will break Laravel 5.4 implementations. I think it's better to use `5.4.*|5.5.*`.

Anyways, will hold off on any changes related to 5.5 and will get back on this once it's officially released.

leyluj:
Noted, Chief!

PR #1221, opened 2017-06-21
[7.0] The global search keywords should not be an array

liub1993:
The Str::lower($value) can not accept array, but sometimes we get array from the client, then exception occurs like:

"error":"Exception Message:\n\nmb_strtolower() expects parameter 1 to be string, array given"


yajra:
Thanks!

PR #1238, opened 2017-07-01
[7.0] Fix multi-column sort of collection. Fix #1237

### LINKED ISSUE #1237: multiple column sorting with collection engine
I've found that if DataTables requests sorting by multiple columns, the returned data is only sorted by the last orderable column when using the collection engine.

Based on this discussion regarding sorting a collection by multiple columns ...

https://stackoverflow.com/questions/25451019/what-is-the-syntax-for-sorting-an-eloquent-collection-by-multiple-columns

I tried replacing the foreach loop in the ordering() function with the code below, and that seems to correctly sort the collection by multiple columns.

```
        $criteria = $this->request->orderableColumns();
        if ($criteria)
        {
            $comparer = function ($first, $second) use ($criteria) {
            foreach ($criteria as $orderable) {
                $column = $this->getColumnName($orderable['column']);
                // normalize sort direction
                $direction = $orderable['direction'];
                if ($first[$column] < $second[$column]) {
                    return $direction === "asc" ? -1 : 1;
                } else if ($first[$column] > $second[$column]) {
                    return $direction === "asc" ? 1 : -1;
                }
            }
            // all elements were equal
            return 0;
            };
            $this->collection = $this->collection->sort($comparer);
        }
```

jond64:

### DIFF
--- a/src/Engines/CollectionEngine.php
+++ b/src/Engines/CollectionEngine.php
@@ -122,20 +122,21 @@ class CollectionEngine extends BaseEngine
             $comparer = function ($a, $b) use ($criteria) {
                 foreach ($criteria as $orderable) {
                     $column = $this->getColumnName($orderable['column']);
-                    if (strnatcasecmp($a[$column], $b[$column]) != 0) {
-                        $direction = $orderable['direction'];
-                        if ($direction === 'desc') {
-                            $first = $b;
-                            $second = $a;
-                        } else {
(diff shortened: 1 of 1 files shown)


yajra:
Will try to test this as soon as I can. Thanks!

yajra:
Released on v7.9.2 👍 thanks!

PR #1255, opened 2017-07-10
Readme: implictly set "--provider" when publishing

uniconstructor:
"--provider" wiil prevent you from accidentally publishing content from other service providers

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,7 +75,7 @@ $ composer require yajra/laravel-datatables-oracle:^7.5
 
 #### Configuration and Assets
 ```bash
-$ php artisan vendor:publish --provider="Yajra\Datatables\DatatablesServiceProvider" --tag=datatables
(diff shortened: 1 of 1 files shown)


yajra:
If we will use the provider flag, then I guess we can remove `--tag=datatables` since it's basically doing the same task.

yajra:
On the other hand, don't you think that it's easier to remember the tags than using the provider?

uniconstructor:
Yes, I think that remove `--tag=datatables` is a good idea.
About tag vs provider option: yes tag is simpler to remember, but DataTables is a very popular plugin and there is a chance that some other Laravel package will use `--tag=datatables` for resource publishing.
To be honest, chances of tag name collision is not very high here, but I prefer to prevent it completely.

yajra:
Good point on possibility of name collision. Thanks!

PR #1265, opened 2017-07-14
Add model instance to blade compiller

VRuzhentsov:
What if i want to work with my model in blade template when added it by `->addColumn()` ? 
For example, to build complex form with some relations called inside.
Or, table of `Projects`, which have `HasOne` relation with `Offers`.
Before, we could create `Offer` instance with `$offer = Offer::find($offer['id])`, which proceed connection to DB.
Now we can simply call `$project->offer`

### DIFF
--- a/src/Helper.php
+++ b/src/Helper.php
@@ -117,6 +117,10 @@ class Helper
      */
     public static function getMixedValue(array $data, $param)
     {
+        if($param instanceof \Illuminate\Database\Eloquent\Model) {
+            $data[lcfirst(class_basename(get_class($param)))] = $param;
+        }
+
         $param = self::castToArray($param);
 
         foreach ($data as $key => $value) {


yajra:
Thanks! But I have done something similar on the [master branch](https://github.com/yajra/laravel-datatables/commit/54705eab211894294b1df27e8fa724fdd96dc08a) and will be released on v8.

Not sure if we can also implement the same on v7 since it's a potential breaking change.

VRuzhentsov:
@yajra That's great, thanks. When will be released v8? and also, anything from my code can be useful to you?

yajra:
@VRuzhentsov, I'm still planning to support Laravel 5.4 on v8 but still trying to balance all the breaking changes I have done on v8 and documenting it. Lots of changes on v8 but the package will be more flexible and I guess more intuitive. 

See https://github.com/yajra/laravel-datatables/blob/master/CHANGELOG.md for some other changes documented. 

With regard to code, I think the dynamic key name would be useful than the hard coded `model` key name.

VRuzhentsov:
@yajra Okay, thanks. I'll close PR, and will be wait for v8.

PR #1293, opened 2017-07-27
[7.0] Fixed configuration override issue.

zgldh:
Fixed configuration override issue.
Otherwise the databases.php configuration will never take effect.

### DIFF
--- a/src/Engines/BaseEngine.php
+++ b/src/Engines/BaseEngine.php
@@ -660,7 +660,7 @@ abstract class BaseEngine implements DataTableEngineContract
         $config  = config('datatables.columns');
         $allowed = ['excess', 'escape', 'raw', 'blacklist', 'whitelist'];
 
-        return array_merge(array_only($config, $allowed), $this->columnDef);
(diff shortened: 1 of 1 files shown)


yajra:
Released on v7.9.6 Thanks!

yajra:
@zgldh can you please test #1311 if it fixes this PR? I have reverted the changes you made because it make XSS protection invalid. Hopefully, #1311 can fix the configuration override issue. Thanks!

yajra:
BTW, I think the issues introduce by this PR are the ff: #1306 #1307 #1305 #1302

lk77:
this issue is solved, 
with 7.9.8 i don't have two times action or *, 
but the problem with escaping still exist,
columns names are stored as value not key in escapeColumns and rawColumns,
and the key does not correspond to the order of the columns.

zgldh:
Thank you @yajra , I think the #1311 update is OK for me.

yajra:
@zgldh thanks for the feedback. Will merge the PR then.

yajra:
Patch released on v7.9.9. Thanks!

PR #1298, opened 2017-07-28
[7.0] Whitelisting src directory in phpunit.xml

lk77:
Hello,

i'm whitelisting src directory in phpunit.xml, because phpunit fails to execute code coverage without it.
i think we should do this on master has well.

thanks.

### DIFF
--- a/phpunit.xml
+++ b/phpunit.xml
@@ -16,7 +16,7 @@
         </testsuite>
     </testsuites>
     <filter>
-	<whitelist processUncoveredFilesFromWhitelist="yes">
+	    <whitelist processUncoveredFilesFromWhitelist="yes">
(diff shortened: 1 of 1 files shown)


phpunit.xml
```diff
@@ -16,6 +16,9 @@
         </testsuite>
     </testsuites>
     <filter>
+	    <whitelist processUncoveredFilesFromWhitelist="yes">
```
yajra: I think there is a missing space for indentation?

yajra:
Looks good. Will merge when the missing space is added.

Please also submit on master branch. Thanks!

yajra:
Released on v7.9.6 Thanks!

Issue #1305, opened 2017-07-31
datatables laravel returned html element as a text


AbuHamdah:
The laravel datatables return an HTML as a text I used a rawColumns and escapeColumns but not working

Please what a solution and thank you

Please find a below pictures
![dara](https://user-images.githubusercontent.com/17378609/28770955-b81689be-75e9-11e7-877a-cf0eacd9fd13.PNG)
![dara2](https://user-images.githubusercontent.com/17378609/28770956-b81c70f4-75e9-11e7-93f5-cf5ddedb9b01.PNG)

### System details

- Operating System
- PHP Version 
- Laravel Version 5.3
- Laravel-Datatables Version 7.9

lk77:
Hello, 

it's normal, escapeColumn is for escaping columns, so it's escaped.
change by 
`->escapeColumns(['id', 'username']`

AbuHamdah:
But, what I should do, to return HTML element? 
Thank you for your response

lk77:
you can use rawColumns(['link'])

AbuHamdah:
I used it, this same problem appeared

lk77:
try to use version 7.9.5, it's working for me, perhaps it's a recent bug.

AbuHamdah:
I used the last version of datatables  laravel, 
I tried also ['link' => true] but not working ,
Must i upgrade laravel version to 5.4 ?

lk77:
i don't think it will solve your issue

AbuHamdah:
What should I do in your opinion?

lk77:
you tried escapeColumns(['link' => true) of rawColumns(['link' => true]) ?

AbuHamdah:
Yes, it is.

lk77:
i mean which one you tested ?

AbuHamdah:
I tested them both

lk77:
escapeColumns([]) should disable escaping on all columns as a workaround

AbuHamdah:
I tried all solutions but nothing be changed

lk77:
and escapeColumns(null) ?

AbuHamdah:
Same problem,
I have question, why the action column return HTML  but the link not?

lk77:
it should work i don't understand.
you removed all the escapeColumns calls ?

AbuHamdah:
How can i remove all escapeColumns calls? and why?

lk77:
inside your code i mean,
[...]

Issue #1302, opened 2017-07-29
Latest 7.x versions has bugs


lorvent:
Hello,
either its just me or for everyone...

latest versions not working.

proper working version for me: 7.8.1

7.9.1 : datatable loading properly but search failing

7.9.3 - 7.9.6 : search not working, removeColumn not working.

is it just me or other people also having this issue?

yajra:
@lorvent can you please paste some snippets how to reproduce the issue? Are you using collection? Anyways, I think a recent PR may have cause an issue and I made a patch for a possible fix. Can you please check again.

lorvent:
Hello Yajra, thanks for quick response

1. yes, using collections

2. it fixes 
* removeColumn
* escapeColumns

**but search problem still exists**

when i enter first character in search box...it shows no results even though results should be shown.

controller snippet

````

$staffs = $this->userRepository->getAllNew()
            ->get()
            ->filter(function ($user) {
                return ($user->inRole('staff') && $user->id!=$this->user->id);
            })->map(function ($user) {
                return [
                    'id' => $user->id,
                    'full_name' => $user->full_name,
                    'email' => $user->email,
                    'created_at' => $user->created_at->format($this->date_format),
                ];
            });

        return $datatables->collection($staffs)
            ->addColumn('actions', 'some buttons')
            ->removeColumn('id')
            ->escapeColumns( [ 'actions' ] )->make();

````

**Comment by lk77** [comment]
escape column not working as well , 
see https://github.com/yajra/laravel-datatables/pull/1306 for the fix.

**Comment by yajra** [comment]
Maybe this is because you are using `->make()` which in return uses the index on the column instead of the name. Can you also please paste your js scripts?

Will also check @lk77 PR later when I got the chance.

**Comment by yajra** [comment]
In addition, since you are using `make()` (array response) you also need to use the index for rawColumns.

```php
->escapeColumns( [ 5 ] )
->make()
```

See http://dt54.yajrabox.com/eloquent/array for ref.

yajra:
Found another issue on config override related to xss and was fixed v7.9.8. Can you please verify if it fixes the issue? Thanks and sorry for the inconvenience.

lorvent:
Hello,

i just tried with 7.9.9 and search problem still exists.

7.8.1 search works properly so comparing codebase with regards to 7.8.1 may give you clue.

thanks.

lorvent:
Hello,
Just bumping the thread.

[...]

PR #1306, opened 2017-07-31
[7.0] Fix escapeColumns bug.

lk77:
Hello,

i'm think this is a bug and not intended.
there is an array_diff_key inside escapeColumns in DataProcessor, but it should be an array_diff i think, because it's the value that matter.

it may be related to https://github.com/yajra/laravel-datatables/issues/1305 and https://github.com/yajra/laravel-datatables/issues/1302

tests are not passing, because they test the old behaviour, so if this gets merged we will need to change the tests.

thanks

### DIFF
--- a/src/Processors/DataProcessor.php
+++ b/src/Processors/DataProcessor.php
@@ -217,7 +217,7 @@ class DataProcessor
             if ($this->escapeColumns == '*') {
                 $row = $this->escapeRow($row);
             } elseif (is_array($this->escapeColumns)) {
-                $columns = array_diff_key($this->escapeColumns, $this->rawColumns);
+                $columns = array_diff($this->escapeColumns, $this->rawColumns);
                 foreach ($columns as $key) {
                     array_set($row, $key, e(array_get($row, $key)));
                 }


yajra:
I think `array_diff_key` is the correct one to use because we are matching keys and not the values. In cases where `make(false)` is used, you then need to use the column index when using `rawColumns` or `escapeColumns` which I think is the issue on #1302.

yajra:
Issue should be fixed on #1310. Thanks!

lk77:
i'm using make(true), and it's not working.
the keys are not present in the array,
just the column names,
like that :
```
[
    0  =>  'action'
]
```

lk77:
i don't know if we need to change how the arrays are populated or to change the check.
btw, `foreach ($columns as $key)` is misleading because $key is the value.

yajra:
@lk77 I just reviewed this and found the issue valid. 

Got confused a bit how this works. This is how it should work as I remember:

Using `escapeColumns` will encode any defined columns except when the column is set as `raw`. 
`rawColumns` are prioritized over `escapeColumns`.

yajra:
Released on v7.9.9 Thanks!

PR #1395, opened 2017-09-16
[6.0] Return true for true, true(string) and 1 in isColumnSearchable method

joeshee:
Validate the key `searchable` in request input against all possible true values `[true, 'true', 1]`

### DIFF
--- a/src/Request.php
+++ b/src/Request.php
@@ -119,10 +119,10 @@ class Request extends IlluminateRequest
     public function isColumnSearchable($i, $column_search = true)
     {
         if ($column_search) {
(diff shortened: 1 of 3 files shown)


yajra:
Looks good but not sure why the test is failing...

joeshee:
Hi, I updated the logic to evaluate true in isColumnSearchable method and it validates all the test cases. But it seems there are other test cases in this branch which failing. 

On debugging I found the problem is at: 
BaseEngine@render on line 693
`return new JsonResponse($output, 200, Config::get('datatables.json.header', []), Config::get('datatables.json.options', 0));
`
I think test cases are unable to mock the Config facade, thus failing.

yajra:
Released on v6.29.1, thanks!

PR #1416, opened 2017-09-29
[8.0] Add addColumns() to EloquentDataTable.

ElfSundae:
It is useful when you want to add many invisible attributes.

```php
$dataTable->addColumns(['foo', 'a' => 'b'], 3);
```

equals:

```php
$dataTable->addColumn('foo', function ($model) {
    return $model->foo;
}, 3);

$dataTable->addColumn('a', function ($model) {
    return $model->b;
}, 4);
```

### DIFF
--- a/src/EloquentDataTable.php
+++ b/src/EloquentDataTable.php
@@ -29,6 +29,28 @@ class EloquentDataTable extends QueryDataTable
     }
 
     /**
+     * Add columns in collection.
+     *
+     * @param  array  $names
+     * @param  bool|int  $order
+     * @return $this
+     */
+    public function addColumns(array $names, $order = false)
+    {
+        foreach ($names as $name => $attribute) {
+            if (is_int($name)) {
+                $name = $attribute;
+            }
+
+            $this->addColumn($name, function ($model) use ($attribute) {
+                return $model->getAttribute($attribute);
+            }, is_int($order) ? $order++ : $order);
+        }
+
+        return $this;
+    }
+
+    /**
      * If column name could not be resolved then use primary key.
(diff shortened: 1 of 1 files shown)


yajra:
Looks good as an alternative of using `$appends` on model. However, do you think it's better to use something like `->addMutators()` instead? This way, we can give a hint on users that this only works if you have some getter attribute on your model?

ElfSundae:
Actually I mean the [hidden attributes](https://laravel.com/docs/5.5/eloquent-serialization#hiding-attributes-from-json), using `$hidden` or `$visible` on model.

For example, I don't want `foo`, `bar` of `User` to be visible in the API result, then I add them in User's $hidden, but I want them to be visible to dataTables which is shown in the admin dashboard.

ElfSundae:
Model attributes in `$hidden` will not be included in the JSON representation, then `$htmlBuilder->columns(['hidden_attributes'])` will be failed to render. You may use `$dataTable->addColumn()` to add these data:
```php
$eloquentDataTable
    ->addColumn('hidden_attribute', function ($model) {
        return $model->hidden_attribute;
    })
    ->addColumn('hidden_attribute1', function ($model) {
        return $model->hidden_attribute1;
    })
    ->addColumn('hidden_attribute2', function ($model) {
        return $model->hidden_attribute2;
    });
```

`addColumns` in this PR is a batch way:

```php
$eloquentDataTable->addColumns(['hidden_attribute', 'hidden_attribute1', 'hidden_attribute2']);
```

yajra:
Released on v8.1.0, thanks!

PR #1438, opened 2017-10-10
(fix): Update API link

dextermb:
Hey,

I noticed that the API reference link was broken, so I've updated it!


yajra:
Thanks, it was pointed to the package php api but I haven't maintained it well. 👍  pointing it to dataTables.js api.

PR #1445, opened 2017-10-13
[8.0] Remove unused provides() from service provider.

ElfSundae:
The default value of `$defer` is `false`, and this behavior will not be changed ever, so we don't need to override it.

And the `provides()` method is useless for a service provider which is not deferred.


yajra:
Please fix the merge conflicts. Thanks!

ElfSundae:
done.

Issue #1294, opened 2017-07-27
Support for custom engines (eg for mongodb)


pimlie:
For my project I am using mongodb through [laravel-mongodb](https://github.com/jenssegers/laravel-mongodb/). Unfortunately I couldnt get laravel-datatables to work with mongodb, as laravel-mongodb requires different syntax in the QueryBuilderEngine as it does not use sql.

To still use mongodb I've written a rough [mongodb engine](https://github.com/pimlie/laravel-datatables-mongodb/) for datatables. But as it seems at the moment there is no easy way of adding a custom engine to datatables, I had to completely exend the laravel-datatables project and e.g. overwrite all methods where configuration settings are pulled.

As this implementation is very ugly and difficult to maintain, do you have any recommendations how to better add a custom engine for mongodb?

yajra:
On version 8, I am also trying to create an engine for Laravel Scout. See https://github.com/yajra/laravel-datatables-scout for reference for possible implementation.

pimlie:
Will you be changing the general engine interface in v8 as well, like removing the `eloquent` and `queryBuilder` methods in `Datatables.php`? From a code perspective I think it would be best to keep the interfaces for built-in engines and custom engines as much the same as possible.

One way to keep the interface more uniform would be by adding a magic __call method like the following pseudo code:
```php
// Datatables.php
public function __call($method, $parameters)
{
  $engines = $config->get('datatables.engines');
  if (!isset($engines[$method])) {
    throw new BadMethodCallException("Method {$method} does not exist.");
  }

  return $engines[$method]::getInstance($parameters);
}
```

Then you would only need to add your engine to the config.engines (and builder) array and you could just do `return Datatables::scout($model);` and/or `return Datatables::of($scoutModel);`.

Only difficulty with this solution i see currently is that e.g. jenssegers/mongodb extends the Eloquent classes, so the `of` method in `Datatables.php` would probably return an Eloquent engine and not a Moloquent engine. Either people should be instructed to add their custom engines definition _before_ the eloquent definition (in `config/datatables.php`) or we should add a check for these kind of inheritances between engines on startup.

yajra:
> Will you be changing the general engine interface in v8 as well, like removing the eloquent and queryBuilder methods in Datatables.php?

The interface would remain the same. The only major breaking change is the namespace and class name being updated from `Datatables` to `DataTables`. 

[...]

PR #1462, opened 2017-10-21
[8.0] Support for plugin engine methods.

pimlie:
As discussed in #1294 

I have implemented more dynamic/flexibel methods for creating DataTable engines instances, the changes are:
- added support for macros in `DataTables.php`
- added `boot` method in `DataTablesServiceProvider` to add macro's for all engine names listed in `config/datatables.php`. Engine names are converted to camelCase, so if you list an engine as `mongodb-query` the macro will be named `mongodbQuery`.
- added `canCreate` and `create` methods in `DataTableAbstract`. 
With the `canCreate` method a DataTableEngine can indicate whether they can be create with the supplied parameters (see Collection for the most interesting use-case).  The `create` method then returns an instance of the DataTable engine. With these 2 methods DataTable Engine plugins have full control over how they should be used. See the CollectionEngine again which required a special treatment in `DataTables.php` (with the conversion from array to Collection) but is now fully self sufficient ;)

Due to the `canCreate` method the `config.datatables.builders` option has become redundant, see the new implementation of `DataTables::make`. Therefore I would like to propose to comment out the builder options in `config/datatables.php` by default and mark that as a 'advanced configuration' option. That way advanced users can still override the builder to engine mapping, but if eg an Engine would split support for one of its supported Builders to a different class the bulk of the users are not affected.


src/CollectionDataTable.php
```diff
@@ -23,6 +23,20 @@ class CollectionDataTable extends DataTableAbstract
      */
     public $original;
 
+    public static function canCreate($source)
+    {
+        return is_array($source) || $source instanceof Collection;
+    }
+
+    public static function create($source)
```
yajra: Please add the missing docs.

src/DataTableAbstract.php
```diff
@@ -137,6 +137,28 @@
     protected $config;
 
     /**
+     * Can the DataTable engine be created for these parameters
```
yajra: Please add all the missing period on docs.

yajra:
PR looks great ❤️ . Need some minor changes and I think this is good to go. Will do some actual testing on a project though when I got the chance just to make sure. But I guess the test is sufficient enough to tell that this works :).

pimlie:
I have added the docs. I actually didnt include them in Collection on purpose because phpdoc should inherit the docs from the Abstract class by default anyway, but I guess this is indeed better as now they list the correct param/return types :)

What are your thoughts about the `query` / `queryBuilder` inconsistency we currently have? The engine for `Query\Builder::class` are called _just_ `Query` but the direct access method in `DataTables.php` is called `queryBuilder`. Renaming to `eloquent-query' would maybe be the best option? E.g I also use mongodb-query in my mongodb [plugin](https://github.com/pimlie/laravel-datatables-mongodb). Then we could deprecate the `QueryDataTables` class and queryBuilder method and remove it in a next major release.

Also, I left in the `eloquent` and `collection` methods in `DataTables.php` but those methods could be removed and replaced by a engine macro. What do you prefer?

yajra:
Actually, queryBuilder is a fluent query builder and not eloquent. Maybe we can add another class `QueryBuilderDataTable` that just extends `QueryDataTable` as a work around? Then we can safely remove it on the next major release?

src/QueryDataTable.php
```diff
@@ -45,6 +45,11 @@ class QueryDataTable extends DataTableAbstract
      */
     protected $limitCallback;
 
+    public static function canCreate($source)
```
yajra: Missing docs. Maybe we can use `@inheritdoc`? But I usually copy and paste the comments :)

src/EloquentDataTable.php
```diff
@@ -15,6 +16,11 @@ class EloquentDataTable extends QueryDataTable
      */
     protected $query;
 
+    public static function canCreate($source)
```
yajra: Missing doc.

pimlie:
@yajra Sorry for missing those phpdocs in the previous commit, I have added them :sweat_smile:

I have also added a commit with my proposal for deprecating the QueryDataTable class, hope you dont mind but it was just as easy as first discussing the required changes here. I have also added a test to make sure we still fully support the old datatables config using the deprecated engine. That test can be removed when we really remove the QueryDataTable class. 

Btw, if you prefer a second PR for that I can rollback that commit but then I have to wait until you accept this PR due to too many files that are changed in both PR's ;)

src/config/datatables.php
```diff
@@ -39,19 +39,21 @@
      * This is where you can register your custom dataTables builder.
      */
     'engines'        => [
-        'eloquent'   => \Yajra\DataTables\EloquentDataTable::class,
-        'query'      => \Yajra\DataTables\QueryDataTable::class,
```
yajra: I think we can retain `query` engine to prevent breaking change. This way, users with old config won't break?

Also, this would also become an additional feature where we can just use `DataTables::query()` for a shorter syntax?

yajra:
I just realized that I think `DataTables::query()` is quite a good shortcut command and I think we can retain it. Although it seems redundant but I think it's quite a good option to have? What do you think?

pimlie:
I dont have a strong preference for either `queryBuilder` or `query`, for me it was more about the inconsistency. But in general shorter names are better I think so maybe it would be better to keep `query` indeed ;)

src/config/datatables.php
pimlie: I added a Deprecated test as well to make sure we still support the old config's. Maybe I missed a use-case in that? See: https://github.com/pimlie/laravel-datatables/blob/feature-support-plugin-engine-methods/tests/Integration/DeprecatedQueryEngineTest.php

yajra:
In my opinion, Yes, we deprecate `queryBuilder` in favor of `query`.

src/config/datatables.php
yajra: Yeah my mistake, I think this is not a breaking change but I think it's nice to have `query()` engine? Thanks!

pimlie:
Great, I have rolled back the Deprecation commit and added a new one. I also renamed the integration test and added a test for the (now) deprecated queryBuilder method just to be sure :)

src/CollectionDataTable.php
```diff
@@ -24,6 +24,32 @@ class CollectionDataTable extends DataTableAbstract
...
+    public static function canCreate($source)
+    {
+        return is_array($source) || $source instanceof Collection;
+    }
+
+    /**
+     * Factory method, create and return an instance for the DataTable engine.
+     *
+     * @param array|\Illuminate\Support\Collection $source
+     * @return \Illuminate\Support\Collection
```
yajra: I think the correct return doc here is `CollectionDataTable|DataTableAbstract`?

src/DataTables.php
```diff
@@ -80,26 +84,35 @@ public function getConfig()
         return app('datatables.config');
     }
 
+
```
yajra: Excess new line?

pimlie: Yes :)

src/CollectionDataTable.php
pimlie: I believe not, the `|` [indicates](https://docs.phpdoc.org/references/phpdoc/tags/return.html) that either of those two types can returned but in this case it always returns a CollectionDataTable (which just happens to be an instance of DataTableAbstract). I actually think you shouldnt use DataTableAbstract as a type anywhere as an abstract class can never be instantiated and its always the parent class you should list. But correct me if I am wrong..

yajra: Not really sure too. Sometimes, scrutinizer complains on things like this. Using PHPStorm, it will suggest to return both so I often go towards it. :)

pimlie: Thats probably because it doesnt understand that with the `parent::create` we call a method of the abstract class which uses the `static` keyword to return an instance of the current class.

The following code will probably trick PHPStorm to get it right:
```php
$datatable = parent::create(source);
/* @var $datatable CollectionDataTable */
return $datatable;
```
Should I use that?

pimlie: Maybe `/* @var $datatable self */` works too btw

yajra: I think it is safe to use `CollectionDataTable|DataTableAbstract`. This way, IDE like phpstorm won't complain since it is compatible on the parent class as it returns `DataTableAbstract`.

pimlie: Yes boss, done and done :smile:

yajra:
Released on 🚀 v8.3.0. ❤️ ❤️ ❤️ Thanks!

pimlie:
Thank you for merging :+1: 

Fyi, for my mongodb plugin I have added another method for easy registration of custom engines by adding a serviceprovider to it. In the serviceprovider I merge the available engines dynamically to the `datatables.php` config, the performance hit of that should be minimal (compared to listing them hardcoded in the config) and makes it so much easier / cleaner

yajra:
Oh yeah, that's a good idea. Will update the scout plugin too when I got the chance. It's not yet stable :)...

BTW, would you mind if I link any mongodb related issues to your plugin? Doesn't have much knowledge on mongodb as I haven't use it yet personally on any project. TIA.

pimlie:
Sure no problem!

ElfSundae:
Oh my...

ElfSundae:
I am confused about the purpose of `DataTables` class for a long time.

In my opinion, DataTables **singleton** may be not needed if there is a static `make` method. Besides, `datatables()` helper always returns a new DataTables instance. And `DataTables::$request` seems unused, and I think if someone wants `request`, `config`, `html`, why not just call `app(...)` to get instance by using container's factory methods.

And, for me, if the engines are configurable, why we define `eloquent()`, `collection()`, `query()` methods in the DataTables? Think about this, I used `DataTables::eloquent()` before to build datatables for my models, now I need some extending of `Yajra\DataTables\EloquentDataTable`, I create a subclass `CustomEloquentDataTable`, then how to let `DataTables::eloquent()` return `CustomEloquentDataTable`? There is no elegant way.

ElfSundae:
And I don't think depending upon `Macroable` is a good design.

pimlie:
I already mentioned in my 2nd post that the `eloquent()` etc methods are redundant now and I think your case is a good reason why we could / should remove them.

But I dont think depending on `Macroable` is really that of a problem, can you explain why you think that? E.g. the problem you describe wouldnt even be a problem if the eloquent method was also a macro based on the configured engine key. Then you could have a `'eloquent' => 'mycustomengine'` record in your config and you are good to go.

That said I agree with you that those methods arent really that necessary, there are already many other ways of creating a datatable engines.

ElfSundae:
I think the Macroable is only useful for the lazy package/module users. 
Inside this package itself, should we really need to store many reduplicate Closures at runtime in `ServiceProvider::boot()`?

pimlie:
So do you have another solution __besides__ just removing the closures for the lazy users?

Btw, why do you say 'reduplicate'?

yajra:
@ElfSundae 

> I am confused about the purpose of DataTables class for a long time.

It was originally inherited from the [original package](https://github.com/bgultekin/laravel4-datatables-package) and was retained for familiarity of users migrating to this package.

> In my opinion, DataTables singleton may be not needed if there is a static make method.

It was added for facade to work. Or I might be wrong?

> DataTables::$request seems unused

It was used on older version of the package and by the buttons plugin. It was retained for backward compatibility. But I guess we can deprecate it and be removed on next major version?

ElfSundae:
@pimlie  I think this PR made DataTables factory more extendable, but with a **complex** API. If someone wants to extend an exist engine or add new, he should be careful, as the doc of `builders` config saying:

> Note, only change this if you know what you are doing!

Before extending, the users may need to know exactly what is the meaning of this doc line. Commented `builders` config, register/replace macros to DataTables, `canCreate()`, `create()`, etc.
Maybe we should explain some extending architecture in the documentation?

"reduplicate" means `DataTables::macro` within `foreach`, registers the same Closures many times.

ElfSundae:
@yajra 
Personally, I prefer instance methods in a singleton, not static methods. Do not mind it, it is all right.

Think about current `datatables()` helper:

```php
use Yajra\DataTables\DataTables;

function datatables($source = null)
{
    if ($source) {
        return DataTables::make($source);
    }

    return new DataTables;
}
```

If someone wants to override `DataTables`, he may replace 'datatables' singleton in a service provider, but `datatables()` helper always use the `Yajra\DataTables\DataTables` class. It can be changed to:

```php
function datatables($source = null)
{
    if ($source) {
        return app('datatables')->make($source);
    }

    return app('datatables');
}
```

Since DataTables is a singleton, letting `make()` become an instance method seems more reasonable, and in the instance `make()` we can access instance variables or methods. 
And after doing this change, we disable `Yajra\DataTables\DataTables::make()`, in my opinion, it is a bad way as it is hard coded the certain class.

ElfSundae:
@pimlie @yajra

Does the following make sense?

**Configuration**

```php
'engines' => [
    'eloquent' => Yajra\DataTables\EloquentDataTable::class,
    'query' => Yajra\DataTables\QueryDataTable::class,
    'collection' => Yajra\DataTables\CollectionDataTable::class,
],

'builders' => [
    'eloquent' => [
        Illuminate\Database\Eloquent\Relations\Relation::class,
        Illuminate\Database\Eloquent\Builder::class,
    ],
    'query' => Illuminate\Database\Query\Builder::class,
    'collection' => Illuminate\Support\Collection::class,
],
```

I exchange key and value for `builders` config.

**DataTables**

```php
namespace Yajra\DataTables;

class DataTables
{
    public function make($source)
    {
        foreach (config('datatables.builders') as $engine => $classes) {
            foreach ((array) $classes as $class) {
                if ($source instanceof $class) {
                    return $this->createDataTable($engine, func_get_args());
                }
            }
        }

        throw new \Exception('No available engine for ' . get_class($source));
    }

    protected function createDataTable($engine, $parameters)
    {
        $class = config('datatables.engines.'.$engine);

        if (! $class) {
            throw new \InvalidArgumentException("DataTable engine {$engine} does not exist.");
        }

        // If we want to support dependency injection for DataTable classes
        // return app()->make($class, $parameters);

        return (new \ReflectionClass($class))->newInstanceArgs($parameters);
    }

    public function __call($engine, $parameters)
    {
        return $this->createDataTable($engine, $parameters);
    }
}
```

If we would like to add `Macroable`, just handle macros in `__call`.

I am not sure whether we need to support variable parameters for `make()`...

**helper.php**

```php
if (! function_exists('datatables')) {
    function datatables($source = null)
    {
        return is_null($source)
            ? app('datatables')
            : call_user_func_array([app('datatables'), 'make'], func_get_args());
    }
}
```

In this PR, `func_get_args()` is used, maybe variable parameters are useful for custom datatables, then fix `datatables()` helper.

**Usage**

```php
datatables($query);
datatables()->eloquent($query);
\DataTables::eloquent($query);
```

**Example of handling mixed source**

```php
class CollectionDataTable extends DataTableAbstract
{
    public function __construct($source)
    {
        $this->collection = $source instanceof Collection
            ? $source : new Collection($source);

        // ...
    }
}
```

**Extending:**

To override built-in engines or add new, just change the config file:

```php
'engines'        => [
    'eloquent'   => App\DataTables\EloquentDataTable::class,
    'mongodb'    => App\DataTables\MongodbDataTable::class,
    // ...
],
```

If you have different engines for the sources extended the same base class, just take note of the order in `builders` config:

```php
'builders' => [
    'mongodb' => [
        Subclass\Of\Illuminate\Database\Eloquent\Relations\Relation::class,
        Illuminate\Database\Eloquent\Builder::class,
        // ...
    ],
    'eloquent' => [
        Illuminate\Database\Eloquent\Relations\Relation::class,
        Illuminate\Database\Eloquent\Builder::class,
    ],
    // ...
],
```

---

In brief, the DataTables factory only depends on the configuration, no matter this package, user application, or other plugin packages.

yajra:
@ElfSundae maybe you can submit another PR and we can discuss this there further. Your proposal makes sense too and I agree with you with regard to `datatables()` helper. The idea of depending the factory on config is also in sync with https://github.com/yajra/laravel-datatables/pull/1462#issuecomment-339649190.

ElfSundae:
OK

pimlie:
Although I have no issue with most of your suggestions, using `ReflectionClass` should imho always be avoided if possible. If you talk about `Macroable` as a bad design due to closures then we should avoid using `ReflectionClass` at all cost.

But what I wanted to do with this PR is to make the configuration _easier_. Why do we _always_ need to configure a mapping between builder classes and engines when probably for most of the users the default builders provided by an engine are ok? Its only adding another point where you can introduce bugs/mistakes.

Also I am not a big fan of deliberate type switching if you can avoid it. Although php is loosely typed I think it can be confusing when the builders config can have both array as a string/class reference elements.

With the implementation in this PR you can already do what you want with regards to registering custom builders to engines. The _only_ issue what this PR didnt address where those engine specific methods in DataTables. I think your suggestion for using `__call` in DataTables for that is a good solution and that it removes the dependency on `Macrorable` is a nice side effect. The rest is just implementation preferences I guess ;)

pimlie:
Oh, and I indeed used `func_get_args()` to provide a really future ready interface for all engines you can think of. Currently all engines only use one `$source`, but who knows what somebody thinks of as an engine and what needs it have. 

So if you really want we could remove the `func_get_args`/`call_user_func_array` but it was a deliberate decision to future proof the custom engine interface :smile:

pimlie:
Thinking about it, actually I would even prefer to go a step further and also remove the default engine list in the configuration as well. Just put them hard coded in `DataTables` or the ServiceProvider and only use the configuration to _override_ or extend default functionality.

ElfSundae:
Let's discuss in https://github.com/yajra/laravel-datatables/pull/1488

PR #1465, opened 2017-10-23
[8.0] Fix exception/warning for PHP 7.2.

CristianDeluxe:
When you use:

`Datatables::of(User::all())->toJson()`

In PHP 7.2 you will get:

`PHP Warning:  count(): Parameter must be an array or an object that implements Countable in yajra/laravel-datatables-oracle/src/CollectionDataTable.php on line 69`

This creates an empty array if get request it's not defined so count() doesn't fail.

### DIFF
--- a/src/CollectionDataTable.php
+++ b/src/CollectionDataTable.php
@@ -65,7 +65,7 @@ class CollectionDataTable extends DataTableAbstract
      */
     public function columnSearch()
     {
-        $columns = $this->request->get('columns') ?: [];
+        $columns = $this->request->get('columns', []);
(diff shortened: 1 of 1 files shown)


src/CollectionDataTable.php
```diff
@@ -65,7 +65,7 @@ public function count()
      */
     public function columnSearch()
     {
-        $columns = $this->request->get('columns');
+        $columns = $this->request->get('columns') ?: [];
```
yajra: Maybe it is better to use the default value arg of get? Thanks!

```php
$this->request->get('columns', []);
```

PR #1473, opened 2017-10-26
[8.0] Fix filtered records total when using filterColumn.

wuwx:
recordsFiltered count value wrong when using filterColumn method


yajra:
Released on v8.3.1, thanks!

PR #1477, opened 2017-10-27
add periods for php docs

jiwom:
PHP Docs Fix by adding periods to sentence.


yajra:
v5 is not supported anymore. Thanks!

PR #1487, opened 2017-10-28
[8.0] Fix datatables() helper

ElfSundae:
As discussed in https://github.com/yajra/laravel-datatables/pull/1462#issuecomment-339976815

- Use container binding to resolve DataTables factory
- Support variable length parameters

### DIFF
--- a/src/helper.php
+++ b/src/helper.php
@@ -1,21 +1,19 @@
 <?php
 
-use Yajra\DataTables\DataTables;
-
 if (! function_exists('datatables')) {
     /**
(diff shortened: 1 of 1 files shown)


ElfSundae:
Updated: don't support variable length parameters.

Variable length parameters for DataTable is a bad idea, and this will make our factory much complex to not work. Think about `new FooDataTable($query, $array)`, how to autoselect a engine?

Maybe someone wants to implement a complex DataTable, then just use `new FooDataTable(...)` or `DataTables::foo(...)`

PR #1543, opened 2017-12-08
[6.0] Fix issue 932

abiodunjames:
I added a fix to take care `mb_strtolower() expects parameter 1 to be string, array given `  error.


yajra:
Released on v6.29.2, thanks!

PR #1553, opened 2017-12-16
[8.0] Remove useless else statements.

carusogabriel:
I've removed the `else`s when we have already returned something :put_litter_in_its_place:

We can also enable `no_useless_else` in `StyleCI`. Should we?


yajra:
@carusogabriel this looks good. Not familiar with `no_useless_else` settings of StlyeCI but if it does automate this step, then I'm in favor of it.

PR #1577, opened 2018-01-14
Master

aliworkshop:
when i'm use filterColumn , the total record count and the filtered record count are the same
i think that was for that in filterColumn function you dont set $this->isFilterApplied as true

this function should be as this
```php
    public function filterColumn($column, callable $callback)
    {
        $this->columnDef['filter'][$column] = ['method' => $callback];
        $this->isFilterApplied = true;
        return $this;
    }
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -15,11 +15,11 @@
   ],
   "require": {
     "php": ">=7.0",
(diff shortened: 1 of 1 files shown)


yajra:
Master already have 5.6 added. Thanks anyways!

PR #1624, opened 2018-02-13
Change Log contract to LoggerInterface.

LEI:
After updating to Laravel 5.6, this message is displayed on error (e.g. wrong column name):

    Class Illuminate\Contracts\Logging\Log does not exist

As described in the [upgrade guide](https://laravel.com/docs/5.6/upgrade#Logging), `Illuminate\Contracts\Logging\Log` should be changed to `Psr\Log\LoggerInterface`.

### DIFF
--- a/src/DataTableAbstract.php
+++ b/src/DataTableAbstract.php
@@ -3,8 +3,8 @@
 namespace Yajra\DataTables;
 
 use Illuminate\Support\Str;
+use Psr\Log\LoggerInterface;
 use Illuminate\Http\JsonResponse;
-use Illuminate\Contracts\Logging\Log;
 use Yajra\DataTables\Utilities\Helper;
 use Illuminate\Support\Traits\Macroable;
 use Yajra\DataTables\Contracts\DataTable;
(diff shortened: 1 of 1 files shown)


yajra:
Is this backward compatible with lower version of Laravel? We may have to do a major bump if not. Haven't updated to 5.6 yet and I can't verify this yet. Thanks!

yajra:
Maybe we should just use the facade and remove the interface. Or resolve it using IoC `app('log')` key?

LEI:
Not tested, but the upgrade note seems to indicate there is no side effects:

> This interface has been removed since this interface was a total duplication of the  `Psr\Log\LoggerInterface`

The `Psr` namespace should be available before 5.6 as well, I'll try to find since when exactly.

LEI:
The [interface signature](https://github.com/laravel/framework/blob/5.5/src/Illuminate/Contracts/Logging/Log.php) hasn't been modified since its introduction on 5.0, except for a (PSR-2?) style change.

Though both `useFiles` and `useDailyFiles` methods are missing from the [PSR interface](https://github.com/php-fig/log/blob/master/Psr/Log/LoggerInterface.php) introduced in 5.6.

I can't find a note about it in the documentation. I tend to think it should be backward compatible with the DataTable usage of the logger.

yajra:
Confirmed still working on 5.5, thanks!

PR #1628, opened 2018-02-15
Check SoftDeletes on HasOne or BelongsTo relations

drahosistvan:
Right now, if the related table uses SoftDeletes, it is not handled via the `performJoin` method.

### DIFF
--- a/src/EloquentDataTable.php
+++ b/src/EloquentDataTable.php
@@ -138,6 +138,7 @@ class EloquentDataTable extends QueryDataTable
     protected function joinEagerLoadedColumn($relation, $relationColumn)
     {
         $table     = '';
+        $deletedAt = false;
         $lastQuery = $this->query;
         foreach (explode('.', $relation) as $eachRelation) {
             $model = $lastQuery->getRelation($eachRelation);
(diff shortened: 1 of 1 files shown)


src/EloquentDataTable.php
```diff
@@ -160,36 +161,48 @@ protected function joinEagerLoadedColumn($relation, $relationColumn)
...
      * Perform join query.
      *
      * @param string $table
      * @param string $foreign
      * @param string $other
+     * @param string $deletedAt
      * @param string $type
      */
-    protected function performJoin($table, $foreign, $other, $type = 'left')
+    protected function performJoin($table, $foreign, $other, $deletedAt = false, $type = 'left')
```
yajra: Please put `deletedAt` as the last argument to prevent possible breaking change.

src/EloquentDataTable.php
```diff
@@ -160,36 +161,48 @@ protected function joinEagerLoadedColumn($relation, $relationColumn)
...
             $lastQuery = $model->getQuery();
         }
 
         return $table . '.' . $relationColumn;
     }
 
+    protected function checkSoftDeletesOnModel($model)
+    {
+        if (in_array('Illuminate\Database\Eloquent\SoftDeletes', class_uses($model))) {
+            return $model->getDeletedAtColumn();
```
yajra: Does this return the fully qualified column name? Can check the codes yet but PR looks good. Thanks!

yajra:
Sorry for late reply. The PR looks good. Will try to test this on actual project next week and possibly add a test case for this scenario.

yajra:
Released on v8.4.2, thanks a lot! 🥂

PR #1668, opened 2018-03-21
[6.0] Added code for generating editor script

karmendra:
This Code change was done long back when I was trying to use editor along with datatables. I know version 8.0 includes dataTables editor but i was wondering if we can implement the same in version 6.0.

How it works is instead of writing javascript for editor in the blade file, user can key in details of editor in an array in getBuilderParameters() function as following:

    /**
     * Get default builder parameters.
     *
     * @return array
     */
    protected function getBuilderParameters()
    {
	    return [
		    'order'      => [[4, "asc"]],
		    'dom'        => '<"row"<"col-md-2"l><"col-md-4"B><""p>><"row"<"col-sm-12"tr>><"row"<"col-sm-5"i><"col-sm-7"p>>',
		    'buttons'    => [
			    'selectAll', 'selectNone', 'export', 'print', 'reset', 'reload',
			    ['extend' => 'create', 'editor' => 'editor'],
			    ['extend' => 'edit', 'editor' => 'editor'],
			    ['extend' => 'selectedSingle', 'text' => 'Duplicate'],
		    ],
		    'select'     => [
			    'style'    => 'multi',
			    'selector' => 'tr',
		    ],
		    'keys'       => [
			    'columns' => ':not(:first-child)',
			    'keys'    => [9],
		    ],
		    'rowId'      => 'id',
		    'responsive' => true,
		    'editor'     => [
			    'variable' => 'editor',
			    'ajax'     => [
				    'url'     => route('user.editor'),
				    'headers' => [
					    'X-CSRF-TOKEN' => '$(\'meta[name="csrf-token"]\') . attr(\'content\')',
				    ],
			    ],
			    'idSrc'    => 'mtdr_id',
			    'fields'   => [
				    ['label' => 'User Name', 'name' => 'name', 'def' => 1, 'type' => 'readonly'],
				    [
					    'label'   => 'User Type', 'name' => 'type',
					    'type'    => 'select',
					    'options' => ['Admin' => 'Admin', 'User' => 'User'],
				    ],
				    [
					    ['label' => 'Email', 'name' => 'email'],
					    [
						    'label'   => 'Enabled', 'name' => 'enabled', 'type' => 'select',
						    'options' => ['n' => 'n', 'y' => 'y'],
					    ],
				    ],
			    ],
		    ],
	    ];
    }

PR #1693, opened 2018-04-05
Fix ambiguous column 'deleted_at'

cuatao:
Fix error ambiguous when two tables are using SoftDeletes.

### DIFF
--- a/src/EloquentDataTable.php
+++ b/src/EloquentDataTable.php
@@ -214,7 +214,7 @@ class EloquentDataTable extends QueryDataTable
         }
 
         if ($deletedAt !== false) {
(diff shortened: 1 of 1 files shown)


vaanholtje:
When can we expect this fix to be released?

yajra:
Duplicate of #1688, can't check test the PR yet. Can you please check and confirm a feedback. Will merge if it passes for you.

vaanholtje:
Tested both and both seem to fix the issue think #1688 is a cleanir fix

cuatao:
@yajra Tested #1688, it's ok. I think #1668 is cleaner, so I closed my PR.

PR #1782, opened 2018-06-27
Fix for relations ordering

vladrub:

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,5 +1,5 @@
 {
-  "name": "yajra/laravel-datatables-oracle",
+  "name": "vladrub/laravel-datatables-oracle",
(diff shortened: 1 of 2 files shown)


yajra:
Can you please provide some snippets to reproduce the issue the PR is addressing? Thanks!

composer.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "yajra/laravel-datatables-oracle",
+  "name": "vladrub/laravel-datatables-oracle",
```
yajra: Thanks but changing the package name will break the package. Please revert this one.

Seb33300:
I think this is already fixed by https://github.com/yajra/laravel-datatables/pull/3238

Issue #1805, opened 2018-07-24
DATATABLES_ERROR shouldn't be by default null


zeyad82:
### Summary of problem or feature request

DATATABLES_ERROR shouldn't be by default null, specially if APP_DEBUG is set to false.

The queries where explicit on production and the APP_DEBUG was set to false even though.
It doesn't make sense to make the default is to show everything, at least with a simple check if APP_DEBUG is set to false, `throw` can be set instead.

yajra:
I think a similar concern was raised before but I haven't had the chance to check it. Can you please send a PR on your proposed solution? Thanks!

yajra:
PR merged and released on v8.9.1 🚀 thanks and sorry for the delay. 🍻

PR #1811, opened 2018-08-02
DATATABLES_ERROR shouldn't be by default null #1805

zeyad82:


yajra:
Maybe we can just set the default config to `throw` or maybe `custom message`?

https://github.com/yajra/laravel-datatables/blob/8.0/src/config/datatables.php#L73

Issue #1789, opened 2018-07-01
Order and filter not working


aline-matos:
### Summary of problem or feature request

Pagination works fine, but it's not ordering or filtering.

![awesomescreenshot-2018-07-01t22-01-08-680z](https://user-images.githubusercontent.com/5360477/42139312-df5240ce-7d61-11e8-9bdb-ae365b424ef9.gif)

### Code snippet of problem

**Route:**
~~~php
Route::get('home/dataTable', 'HomeController@dataTable');
~~~

**Controller:**
~~~php
public function dataTable(Request $request)
{
    return datatables()->of(Equipments::all())->toJson();
}
~~~

**Javascript:**
~~~javascript
$('.dataTable').dataTable({
        'lengthMenu': [[10, 25, 50, 100, -1], [10, 25, 50, 100, 'Todos']],
        'processing': true,
        'serverSide': true,
        'ajax': 'home/dataTable',
        'columns': [
                { 'data': 'serial_number' },
                { 'data': 'modem_sn' },
                { 'data': 'power_supply_sn' },
                { 'data': 'name' },
                { 'data': 'equipment_type_id' },
                {
                    'data': null,
                    render: function ( data ) {
                        return (data.enabled)
                            ? '<i class="mdi mdi-circle text-success"></i> Online'
                            : '<i class="mdi mdi-circle text-danger"></i> Offline';
                    }
                }
            ],
        'language': {
            'url': 'plugins/datatables-language/Portuguese-Brasil.json'
        }
    })
~~~

**JSON Response test 1:**
~~~json
order: [{column: "0", dir: "asc"}]
~~~

**JSON Response test 2:**
~~~json
order: [{column: "0", dir: "desc"}]
~~~

### System details

- Ubuntu 18.04 LTS
- PHP 7.2.7-1
- Laravel 5.6
- PostgreSQL 9.5.13

yajra:
That's odd. Can you try inspecting the ajax request with `APP_DEBUG=true` and review the input values submitted? It seems like the javascript is fixed on ordering with column 0.

aline-matos:
**Post 1:**
~~~javascript
draw: 1
columns[0][data]: serial_number
columns[0][name]: serial_number
columns[0][searchable]: true
columns[0][orderable]: true
columns[0][search][value]: 
columns[0][search][regex]: false
columns[1][data]: modem_sn
columns[1][name]: 
columns[1][searchable]: true
columns[1][orderable]: true
[...]

PR #1813, opened 2018-08-03
[8.0] Allow boolean values for column[i].searchable

sgotre:
When dt Request is sent as application/json instead of multipart/form-data
and when in ColumnDef boolean values are used the search does not work. 

Did not work
column[i].searchable = true;

Did work
column[i].searchable = 'true';

But the type of searchable is boolean according to 
https://datatables.net/reference/option/columns.searchable

might affect https://github.com/yajra/laravel-datatables/issues/1789

### DIFF
--- a/src/Utilities/Request.php
+++ b/src/Utilities/Request.php
@@ -154,10 +154,20 @@ class Request
     public function isColumnSearchable($i, $column_search = true)
     {
         if ($column_search) {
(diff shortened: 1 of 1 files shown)


yajra:
Wew, sorry for overlooking this PR for so long. Looks good. Thanks!

yajra:
Released on v8 and v9. Thanks!

Issue #1824, opened 2018-08-24
laravel 5.7 support


plindsay:
Hi there, is it possible to support Laravel 5.7?

yajra:
I think 5.7 is not yet officially released? Anyways, sure will add initial support on master branch for 5.7.

yajra:
Initial support added but it's not yet tested. Will try to check it when I got the chance. Please give feedbacks on how it works for you. Thanks!

plindsay:
Great thanks :-) How is it best to test this with Laravel 5.7 via composer?

yajra:
You can require it on composer using `dev-master` version and test the app as usual :)

Rux77:
Laravel 5.7 has been officially released. 
Is it possible to support Laravel 5.7 on the official branch?

hailwood:
@yajra the upgrade guide for 5.7 is out, reading through it, it appears there's nothing in there that would break with this package so testing is pretty good.

Could you please tag a new release?

yajra:
@hailwood yes, Im in the process of testing Laravel 5.7 and updating my packages to support it. I started on my [oracle db](https://github.com/yajra/laravel-oci8) packages and going to review dataTables packages later today.

yajra:
Support released on v8.8.0, thanks!

Jimmy-JS:
hi, @yajra I can't use v8.8.0. even in my composer.json I directly use v.8.8.0 like this:
`"yajra/laravel-datatables-oracle": "8.8.0",`
I got this error
[...]

PR #1830, opened 2018-09-04
Support Laravel 5.7

richardkeep:


drbyte:
In the interim, just use `dev-master` as per #1824 

(Note: That means this PR probably won't get merged, since all that's needed is to tag a new release from the main branch, since these changes are already in `master`.)

yajra:
Added on v8.8.0, thanks!

PR #1852, opened 2018-09-22
[8.0] Only escape callable output of add and edit column.

### LINKED ISSUE #1849: Raw columns
By default blade strings/templates, should be considered as `rawColumn` cause we do the filtering inside of the blade string/template using `{{ }}` operator.

It is not robust to add also a `rawColumn` for each of your columns that use blade.

NOTE: this won't be a breaking-change or cause a security issue, as everyone uses `{{ }}` inside of the blade strings/templates.

sharifzadesina:
Fixes #1849
P.S. bro I think you need to rewrite whole package again, you use nesting extremely.

### DIFF
--- a/src/Processors/DataProcessor.php
+++ b/src/Processors/DataProcessor.php
@@ -100,11 +100,11 @@ class DataProcessor
         foreach ($this->results as $row) {
             $data           = $this->escapeRow(Helper::convertToArray($row));
             $value          = $this->addColumns($data, $row);
-            $value          = $this->addIndexColumn($value, $row);
             $value          = $this->editColumns($value, $row);
             $value          = $this->setupRowVariables($value, $row);
             $value          = $this->selectOnlyNeededColumns($value);
             $value          = $this->removeExcessColumns($value);
+            $value          = $this->addIndexColumn($value, $row);
             $this->output[] = $object ? $value : $this->flatten($value);
         }
 
--- a/src/Processors/DataProcessor.php
+++ b/src/Processors/DataProcessor.php
@@ -104,7 +104,7 @@ class DataProcessor
             $value          = $this->setupRowVariables($value, $row);
             $value          = $this->selectOnlyNeededColumns($value);
             $value          = $this->removeExcessColumns($value);
-            $value          = $this->addIndexColumn($value, $row);
+            $value          = $this->addIndexColumn($value);
             $this->output[] = $object ? $value : $this->flatten($value);
         }
 
@@ -132,10 +132,9 @@ class DataProcessor
      * Process add index column.
      *
      * @param mixed $data
-     * @param mixed $row
      * @return array
      */
-    protected function addIndexColumn($data, $row)
+    protected function addIndexColumn($data)
     {
         if ($this->includeIndex) {
             $data[$indexColumn] = ++$this->start;
--- a/src/Processors/DataProcessor.php
+++ b/src/Processors/DataProcessor.php
@@ -100,7 +100,7 @@ class DataProcessor
     public function process($object = false)
     {
         $this->output      = [];
-        $this->indexColumn = config('datatables.index_column', 'DT_RowIndex');
+        $this->indexColumn = config('datatables.index_column', 'DT_Row_Index');
 
         foreach ($this->results as $row) {
             $data           = $this->escapeRow(Helper::convertToArray($row));
--- a/src/Processors/DataProcessor.php
+++ b/src/Processors/DataProcessor.php
@@ -213,23 +213,23 @@ class DataProcessor
     /**
      * Escape column values as declared.
      *
-     * @param array $output
+     * @param array $row
      * @return array
      */
-    protected function escapeColumns(array $output)
+    protected function escapeColumns(array $row)
     {
-        return array_map(function ($row) {
-            if ($this->escapeColumns == '*') {
-                $row = $this->escapeRow($row);
-            } elseif (is_array($this->escapeColumns)) {
-                $columns = array_diff($this->escapeColumns, $this->rawColumns);
-                foreach ($columns as $key) {
-                    array_set($row, $key, e(array_get($row, $key)));
-                }
-            }
+        if ($this->escapeColumns == '*') {
+            return $this->escapeRow($row);
+        }
 
-            return $row;
-        }, $output);
+        if (is_array($this->escapeColumns)) {
+            $columns = array_diff($this->escapeColumns, $this->rawColumns);
+            foreach ($columns as $key) {
+                array_set($row, $key, e(array_get($row, $key)));
+            }
+        }
+
+        return $row;
     }
 
     /**
--- a/src/Processors/DataProcessor.php
+++ b/src/Processors/DataProcessor.php
@@ -260,15 +260,18 @@ class DataProcessor
 
     /**
      * Whether to escape column or no.
+     *
      * @param string $key
      * @return bool
      */
     protected function shouldEscapeColumn($key)
     {
         if ($this->escapeColumns === '*') {
-            return ! in_array($key, $this->rawColumns); // escape if now a raw column
+            return ! in_array($key, $this->rawColumns); // escape if is not a raw column
         } elseif (is_array($this->escapeColumns)) {
-            return ! in_array($key, array_diff($this->escapeColumns, $this->rawColumns));
+            return in_array($key, array_diff($this->escapeColumns, $this->rawColumns));
+        } else {
+            return false;
         }
     }
 }
--- a/src/Processors/DataProcessor.php
+++ b/src/Processors/DataProcessor.php
@@ -92,20 +92,16 @@ class DataProcessor
         $indexColumn  = config('datatables.index_column', 'DT_Row_Index');
 
         foreach ($this->results as $row) {
-            $data  = Helper::convertToArray($row);
+            $data  = $this->escapeColumns(Helper::convertToArray($row));
             $value = $this->addColumns($data, $row);
+            $value = $this->addIndexColumn($value, $row);
             $value = $this->editColumns($value, $row);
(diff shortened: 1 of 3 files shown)


yajra:
Thanks, can you please fix the failing tests?

> you use nesting extremely

Yeah, will try to refactor more when I got the chance. :)

sharifzadesina:
@yajra Done.
About "extremely nesting" I mean, "code extremely" in the whole of the project.
for example this function [here](https://github.com/yajra/laravel-datatables/blob/8.0/src/Processors/DataProcessor.php#L207-L221).
you can combine it with `escapeRow` and make code cleaner.

sharifzadesina:
@yajra I doubt if they are cause of my code.

yajra:
Just tested this with my app and it breaks my DT with the same error on the tests. Please fix. Thanks!

```php
exception: "Symfony\Component\Debug\Exception\FatalThrowableError"
file: "/www/yajra/dataTables/core/src/Processors/DataProcessor.php"
line: 241
message: "Argument 1 passed to Yajra\DataTables\Processors\DataProcessor::escapeRow() must be of the type array, integer given, called in /www/yajra/dataTables/core/src/Processors/DataProcessor.php on line 223"
```

sharifzadesina:
@yajra Bro I fixed the error, and improved the package a lot.
please test the PR and merge it.
this will make the life for developers easier.
also I hope you will work on the package and improve other parts too.
thank you

yajra:
@sharifzadesina thanks but it's failing on the indexColumn. I think it should be the last item on row manipulations as per the original code before PR.

src/Processors/DataProcessor.php
```diff
@@ -98,21 +98,17 @@ public function process($object = false)
...
-            $value = $this->selectOnlyNeededColumns($value);
-            $value = $this->removeExcessColumns($value);
-
-            if ($this->includeIndex) {
-                $value[$indexColumn] = ++$this->start;
-            }
-
+            $data           = $this->escapeRow(Helper::convertToArray($row));
+            $value          = $this->addColumns($data, $row);
+            $value          = $this->addIndexColumn($value, $row);
```
yajra: I think this should be the last item in row processing or maybe second to the last then escape the row.

sharifzadesina: We only escape row columns here cause we need to escape addon columns on their own function, so we can decide to not escape blade strings.

sharifzadesina:
@yajra I did that, but it is still failing, if you check commits it started failing when I merged `selectOnlyNeededColumns`, and it fails at `filteredRecourds` part. I don't know why cause it is not related to my changes?

yajra:
The changes looks good. However, there are some columns on my project that should be raw but are being escaped. It's not blade but a closure that returns a view.

```php
->rawColumns(['filename_html', 'deleted_html', 'cooked_html']);
```

image


Need to review this further. I think we are missing some tests on raw columns.

sharifzadesina:
@yajra This should be fixed now.

yajra:
Still not working and I think I found where the issue was:

1. escapeRow should be last one to be called since added & edited columns are not yet processed.

```php
        foreach ($this->results as $row) {
            $data  = Helper::convertToArray($row);
            $value = $this->addColumns($data, $row);
            $value = $this->editColumns($value, $row);
            $value = $this->setupRowVariables($value, $row);
            $value = $this->selectOnlyNeededColumns($value);
            $value = $this->removeExcessColumns($value);
            $value = $this->addIndexColumn($value);
            $value = $this->escapeRow($value);

            $this->output[] = $object ? $value : $this->flatten($value);
        }
```

2. Since our goal is to not escape blade string from add/edit column, we can modify our checker to something like:

```php
    protected function shouldEscapeColumn($key)
    {
        foreach ($this->appendColumns as $column) {
            if ($column['name'] == $key && is_string($column['content'])) {
                return false;
            }
        }

        foreach ($this->editColumns as $column) {
            if ($column['name'] == $key && is_string($column['content'])) {
                return false;
            }
        }

        if ($this->escapeColumns === '*') {
            return ! in_array($key, $this->rawColumns); // escape if is not a raw column
        } elseif (is_array($this->escapeColumns)) {
            return in_array($key, array_diff($this->escapeColumns, $this->rawColumns));
        } else {
            return true;
        }
    }
```

src/Processors/DataProcessor.php
```diff
@@ -125,13 +126,28 @@ public function process($object = false)
     protected function addColumns($data, $row)
     {
         foreach ($this->appendColumns as $key => $value) {
-            $value['content'] = Helper::compileContent($value['content'], $data, $row);
+            $value['content'] = Helper::compileContent($value['content'], $data, $row, $this->shouldEscapeColumn($key));
```
yajra: I think the escape flag here is not needed since escaping will be done on `escapeRow` process.

src/Processors/DataProcessor.php
```diff
@@ -142,7 +158,7 @@ protected function addColumns($data, $row)
     protected function editColumns($data, $row)
     {
         foreach ($this->editColumns as $key => $value) {
-            $value['content'] = Helper::compileContent($value['content'], $data, $row);
+            $value['content'] = Helper::compileContent($value['content'], $data, $row, $this->shouldEscapeColumn($key));
```
yajra: Not needed like on `addColumns`.

src/Utilities/Helper.php
```diff
@@ -54,17 +54,18 @@ protected static function isItemOrderInvalid($item, $array)
...
      * @return mixed
      */
-    public static function compileContent($content, array $data, $param)
+    public static function compileContent($content, array $data, $param, $escape = true)
     {
         if (is_string($content)) {
             return static::compileBlade($content, static::getMixedValue($data, $param));
         } elseif (is_callable($content)) {
-            return $content($param);
+            return $escape ? e($content($param)) : $content($param);
```
yajra: Should be reverted back to original since all data here is callable.

src/Processors/DataProcessor.php
```diff
@@ -261,4 +257,18 @@ protected function escapeRow(array $row)
...
     }
+
+    /**
+     * Whether to escape column or no.
+     * @param string $key
+     * @return bool
+     */
+    protected function shouldEscapeColumn($key)
+    {
+        if ($this->escapeColumns === '*') {
```
yajra: We can add the logic here to figure out if added/edited columns are to be escaped.

```php
        foreach ($this->appendColumns as $column) {
            if ($column['name'] == $key && is_string($column['content'])) {
                return false;
            }
        }

        foreach ($this->editColumns as $column) {
            if ($column['name'] == $key && is_string($column['content'])) {
                return false;
            }
        }
```

sharifzadesina:
@yajra we are not doing escaping at "escapeRow", cause we don't know the content was a blade string or a callable, we need to escape output of callables, that's why we only escape database data using `escapeRow`.

yajra:
@sharifzadesina we can identify the content by checking the append columns structure:

```php
foreach ($this->appendColumns as $column) {
            if ($column['name'] == $key && is_string($column['content'])) {
                return false;
            }
        }
```

> that's why we only escape database data

Good point on this one.

yajra:
But I think we should still escape added/edited columns since we can still directly return a database value from a closure.

sharifzadesina:
@yajra
Imagine I've added this callable using `addColumn`:
```php
function () {
    return '<b>XSS</b>';
}
```
The `content` will be string and we won't escape it. this should not happen.
---
Can you please send your full code? I think it should work with the changes I made.

yajra:
For closure like that, I would still prefer to be explicit and add it on the `rawColumns`.

For a straight blade string, It's ok to be raw.
```php
addColumn('xss', function () {
    return '<b>XSS</b>'; // should be added to rawColumns
})

addColumn('xss', '<a href="#">{{ $xss }}</a>'); // can bypass escape
```

yajra:
Already created the [gist](https://gist.github.com/yajra/0163c0622de2c0ba306dedec3379cfda). 

Yes, we should by default escape all rows for xss protection unless explicitly set as raw column.That is why I agreed of allowing `blade templates` to be raw since the argument is we are using `{{ }}` on our template. It's a different case for closure though.

sharifzadesina:
@yajra no I mean the script you use to test my PR

yajra:
It's just a standard dataTable with multiple addColumns. Anyways, I added it on the gist.

sharifzadesina:
@yajra Bro I tested the PR on my own apps, it works without problem, are you sure you are using whole pull request?
cause there were also some changes at `Helper.php` file.

yajra:
It's still failing the last time I checked. For now I think the safest implementation would be:

1. Allow raw columns on the following condition:
- String template

```php
->addColumn('extra', 'some string')
->editColumn('extra', 'some string')
```

- Blade template

```php
->addColumn('extra', 'path.to.blade')
->editColumn('extra', 'path.to.blade')
```

2. Do not allow raw columns on closure:

```php
->addColumn('extra', function($model) {
   return 'html';
})
->rawColumns(['extra'])
```

sharifzadesina:
@yajra It is exactly like that, that's why I told "this is not a BC".

yajra:
It seems to work well now but it breaks the index column. Reverting, please resubmit the PR and fix the index column part. Thanks!

yajra:
Or you can alternatively send another PR to fix the index column? Will hold the revert and release of this feature until it's fixed. Thanks!

yajra:
Fixed the index column and added some tests. Released on v8.13.0 🚀 

If you can, please also submit a PR on [docs](https://github.com/yajra/laravel-datatables-docs). Thanks!

oriceon:
@yajra what`s wrong on last version (rawColumns) with rawColumns() because now it displays columns without escape.

yajra:
Resume discussion on https://github.com/yajra/laravel-datatables/issues/1916.

PR #1882, opened 2018-10-17
[BC] Fix the default name of index column to follow DT syntax.

sharifzadesina:
This is a small PR that changes the name of the index column to follow DT syntax.
DT syntax for special columns is like: `DT_NameName`
But the name of the index column was `DT_Name_Name`
I don't know this is a breaking change or no cause I only changed the default value.

### DIFF
--- a/src/CollectionDataTable.php
+++ b/src/CollectionDataTable.php
@@ -189,14 +189,14 @@ class CollectionDataTable extends DataTableAbstract
     }
 
     /**
-     * Revert transformed DT_Row_Index back to it's original values.
+     * Revert transformed DT_RowIndex back to it's original values.
      *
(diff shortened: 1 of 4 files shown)


yajra:
I think this is breaking change but necessary since I think we can consider this as a bug fix. 

On the other hand, It will not break only if they have published the config. Or just revert the old name by publishing and editing the value on config file.

yajra:
BTW, I think this should also be implemented on https://github.com/yajra/laravel-datatables-html/blob/3.0/src/Html/Builder.php#L472.

yajra:
Will wait until weekend to see if some users have comments on this. If no reactions, I'm ok to release this next week. Thanks!

PR #1926, opened 2018-11-27
Fixes issue for ordering by a relationship.

Nathanw:
This fixes an issue when ordering by a `HasMany` relationship due to no selects being on the query and overriding the original models id with the id from the relationship.

### DIFF
--- a/src/QueryDataTable.php
+++ b/src/QueryDataTable.php
@@ -619,10 +619,7 @@ class QueryDataTable extends DataTableAbstract
                 $column = $this->resolveRelationColumn($orderable['name']);
                 
                 if (str_contains($column, '.')) {
-                    $this->query->addSelect([
-                        $this->query->getModel()->getTable() . '.*',
-                        $column
-                    ]);
+                    $this->applySelects([$column]);
                 }
 
                 if ($this->hasOrderColumn($column)) {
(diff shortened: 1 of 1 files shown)


src/QueryDataTable.php
```diff
@@ -629,6 +633,29 @@ protected function defaultOrdering()
             });
     }
 
+    /**
+     * Apply selects to query.
+     *
+     * @param array $selects
+     */
+    public function applySelects(array $selects)
```
yajra: Please use `protected function`. Thanks!

yajra:
The PR looks good and tests are passing. Will check this asap.

BTW, if possible, can you add a test for this scenario? Thanks! 🍻

yajra:
@Nathanw I just realized after testing this that allowing sort on `HasMany` relationship would yield to multiple duplicate records since we are not grouping on sql level. 

Duplicate records when sorted via User -> HasMany-> Post
image

On the other hand, I think it does fix sorting on HasMany relationship.

Do you think we should merge this and allow the unnecessary duplicate results? 

Any feedbacks from other users?

OzanKurt:
@yajra Almost every relation cause duplicate results. It's always correct to `groupBy` the query inside your `DataTable` class manually. I encountered this problem over 100 times.

Nowadays I always include `$query->groupBy('my_table');` just before I return the results.

Seb33300:
I think this is already fixed by https://github.com/yajra/laravel-datatables/pull/3238

PR #1954, opened 2019-01-07
add ability to display model hidden fields

mo7amed-3bdalla7:
sometime we have a model with hidden array to prevent fields from appear in API but we need it in WEB 
` protected $hidden = [
        'active',
        'updated_at',
        'created_at',
    ];` 

If we need to display fields like  active and created_at we can use
`Datatables::of(MainCategory::query())->make(true, ['active','created_at']);`


yajra:
We already have something similar. See https://yajrabox.com/docs/laravel-datatables/master/add-columns#blade. Thanks!

Issue #2044, opened 2019-04-07
Pagination - count - performance issue


MajkieS:
Hey there,

when having 0,5 million records in a database (everything optimized using query builder instead of collections, using indexes and so) table is blazing fast except during filtration.

When filtering through table records it counts on every request like so: select count(*) as aggregate from.

In file: QueryDataTable.php there is a method: public function count(). This should probably respect option skipTotalRecords and everything can be solved. 

To avoid showing 0 pages in pagination it should be:
if ($this->skipTotalRecords) return null;

Tested, works flawlessly after this small change.

Thanks for feedback.

yajra:
hi @MajkieS, thanks for the inputs. Can you please submit a PR on your proposed solution so that we can check and verify it?

MajkieS:
Hey @yajra, I'll submit a PR.

MajkieS:
@yajra It's done. You can check it out :)

karmendra:
Any update on this?

PR #2050, opened 2019-04-10
[Performance] Skip count method if skipTotalRecords is true

MajkieS:
Related to issue #2044

### DIFF
--- a/src/QueryDataTable.php
+++ b/src/QueryDataTable.php
@@ -186,6 +186,10 @@ class QueryDataTable extends DataTableAbstract
      */
     public function count()
     {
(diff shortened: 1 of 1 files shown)


yajra:
Sorry for late response. As per docs https://yajrabox.com/docs/laravel-datatables/master/skip-total-records:

> Skip total records aims to improve dataTables response time by skipping the total records count query and settings its value equals to the filtered total records.

Adding this PR will change that behaviour. Maybe what we can do is to use https://yajrabox.com/docs/laravel-datatables/master/set-filtered-records?

Issue #2045, opened 2019-04-08
Sorting nested relation that refer to the same table


Morinohtar:
### Issue

Sorting nested relation that refers to the same table.
I tried some solutions that i found here and some others but i'm stuck...

### Code

The function that renders the code is pretty much this:

```php
$foodGroups = FoodGroup::with([
            'parent' => function ($query) use ($locale) {
                $query->select(['id', "name->{$locale} AS nameLocale"]);
            }
        ])
        ->select([
            'food_groups.id',
            'food_groups.parent_id',
            "food_groups.name->{$locale} AS nameLocale",
            'food_groups.published_at',
            'food_groups.created_at',
        ]);

        return DataTables::of($foodGroups)
            ->filterColumn('nameLocale', function ($query, $keyword) use ($locale) {
                $sql = "LOWER(json_unquote(json_extract(`food_groups`.`name`, '$.\"{$locale}\"')))  LIKE ?";

                $query->whereRaw($sql, ["%{$keyword}%"]);
            })
            ->filterColumn('parent.nameLocale', function ($query, $keyword) use ($locale) {
                $query->whereHas('parent', function ($query) use ($keyword, $locale) {
                    $sql = "LOWER(json_unquote(json_extract(`laravel_reserved_0`.`name`, '$.\"{$locale}\"')))  LIKE ?";

                    $query->whereRaw($sql, ["%{$keyword}%"]);
                });
            })
            ->addColumn('checkbox', function ($foodGroup) {
                return view('backend.administrators.pages.food-groups._partials.datatable.checkbox', [
                    'foodGroup' => $foodGroup,
                ]);
            })
            ->editColumn('nameLocale', function ($foodGroup) {
                return view('backend.administrators.pages.food-groups._partials.datatable.name', [
                    'foodGroup' => $foodGroup,
                ]);
            })
            ->editColumn('parent', function ($foodGroup) {
                return view('backend.administrators.pages.food-groups._partials.datatable.parent', [
                    'foodGroup' => $foodGroup,
                ]);
            })
            ->editColumn('published_at', function ($foodGroup) {
                return view('backend.administrators.pages.food-groups._partials.datatable.published-at', [
                    'foodGroup' => $foodGroup,
                ]);
            })
            ->editColumn('created_at', function ($foodGroup) {
                return view('backend.administrators.pages.food-groups._partials.datatable.created-at', [
                    'foodGroup' => $foodGroup,
                ]);
            })
[...]

Issue #2024, opened 2019-03-17
Filter Nested relationship unknown column , Laravel Prefix Table problem


kouloughli-hemza:
### Prefix problem not yet fixed in relationship
i created custom dropdown to filter using aganecy name;
in global search it was working but when i added custom filter i get the following error
SQLSTATE[42S22]: Column not found: 1054 Unknown column 'table.name' in 'where clause'
in matter of sql it expected error because Yajra not adding PREFIX to this table,
not sure why , only in where clause the table not getting prefixed
```php            
$data =  $model->with([
                'Table',
                'Table1.relation_1',
                'Table1.relation_2',
                'Table1.relation_3'
            ])->orderBy('table0.id','desc')->select();
            return $data;
```

This is the generated Sql query : 

```sql
select * from `prefix_table0` left join `prefix_table1` on `prefix_table0`.`id_application` = `prefix_table1`.`id` left join `prefix_table2` on `prefix_table1`.`created_by_agency` = `prefix_tbale2`.`id` where `prefix_table1`.`deleted_at` is null and LOWER(table2.name) REGEXP 88 order by `dz_table0`.`id` desc
```
as you can see the table2 is not getting prefixed ,
this is commun problem in yajra 
im not sure if there is a solution to this

### System details

- Operating System: ubuntu
- PHP Version : 7.2
- Laravel Version : Laravel Framework 5.6.11

yajra:
It seems like you are using a self join relationship and it is currently not supported by the package. You need to use join statements for this to work on self joins.

yajra:
Fixed via #2051.

Issue #1977, opened 2019-02-03
Sorting on nested relation


chancezeus:
### Summary of problem or feature request

When having a model like:

```php
class NestedModel extends Model {
  public function parent()
  {
    return $this->belongsTo(NestedModel::class, 'parent_id');
  }
}
```

If the relation is eager loaded and sorted on, it'll fail with a duplicate table SQL error

### Code snippet of problem

See snippet above

### System details

Latest Laravel Homestead
Latest (released) version of Laravel (5.7.19)
Latest (released) versions of DataTables

### Quickfix
If I place the following code at line 180 of EloquentDataTable.php it'll work (replacing the original performJoin on that line):

```php
            if ($table === $this->query->getModel()->getTable()) {
                $foreign = preg_replace("#^{$table}\\.#", "__nested__{$table}__.", $foreign);
                $deletedAt = $deletedAt ?
                    preg_replace("#^{$table}\\.#", "__nested__{$table}__.", $deletedAt) :
                    $deletedAt;

                $this->performJoin("{$table} as __nested__{$table}__", $foreign, $other, $deletedAt);

                $table = "__nested__{$table}__";
            } else {
                $this->performJoin($table, $foreign, $other, $deletedAt);
            }
```

It's a quick and dirty solution that "works for me", maybe someone can create a more future proof variant, since I didn't check all possible variations (other relation types mostly, this works for BelongsTo, but might fail for the other supported types).

yajra:
Some related links for ref:

https://github.com/yajra/laravel-datatables/issues/880
https://github.com/yajra/laravel-datatables/issues/577
https://github.com/yajra/laravel-datatables/issues/522

yajra:
Fixed via #2051.

Issue #880, opened 2016-12-08
[Bug] Error when searching self join relationships.


coolflames:
### Summary of problem or feature request

I have a problem when search data

### Code snippet of problem
My Model

```php
class Appl_menu extends Model
{
    protected $table = 'appl_menus';
    public $timestamps = false;

    protected $fillable = ['menu_id', 'menu_name', 'menu_module_id', 'menu_file_name', 'menu_id_parent',
        'menu_seqn', 'menu_enable_sts', 'created_by', 'created_date', 'updated_by', 'updated_date'];

    protected $primaryKey = 'menu_id';

    public function nullRelation()
    {
        return null;
    }

    public function menus ()
    {
        return $this->belongsTo(Appl_menu::class,'menu_id_parent');
    }

    public function menuss ()
    {
        return $this->hasMany(Appl_menu::class,'menu_id_parent','menu_id');
    }
}
```
my controller

```php
public function index(Request $request, Builder $htmlBuilder)
    {
        if ($request->ajax()){
            $menuss = Appl_menu::with('modules', 'menus');
            //->orderBy('module_id');
            return Datatables::of($menuss)
                ->addColumn('action', function ($menu) {
                    return view('datatable._action', [
                        'model' => $menu,
                        'form_url' => route('menus.destroy', $menu->menu_id),
                        'edit_url' => route('menus.edit', $menu->menu_id),
                        'confirm_message' => 'Yakin mau menghapus menu ' . $menu->menu_name . '?'
                    ]);
                })->make(true);
        }

        $html = $htmlBuilder
            ->addColumn(['data' => 'menu_id', 'name' => 'menu_id', 'title' => 'ID'])
            ->addColumn(['data' => 'menu_name', 'name' => 'menu_name', 'title' => 'Nama Menu'])
            ->addColumn(['data' => 'modules.module_name', 'name' => 'modules.module_name', 'title' => 'Module'])
            ->addColumn(['data' => 'menus.menu_name', 'name' => 'menus.menu_name', 'title' => 'Parent', 'defaultContent' => ''])
            ->addColumn(['data' => 'action', 'name' => 'action', 'title' => '', 'orderable' => false, 'searchable' => false]);

        return view('menus.index')->with(compact('html'));
    }
```
Error output when search

```
SQLSTATE[42601]: Syntax error: 7 ERROR: syntax error at or near "as"
LINE 1: ...ST("appl_menus" as "laravel_reserved_0.menu_name" as TEXT)) ...
[...]

PR #2051, opened 2019-04-10
[9.0] Sorting nested relation that refers to the same table.

Morinohtar:
A fix for the issue #2045 . It only addresses the "BelongsTo". I did not touch the rest of the relations, so a more complete "fix" can be made.

I also replaced the deprecated function "assertContains" with "assertStringContainsString" in the QueryDataTableTest.

I ran the tests and nothing exploded :P

### DIFF
--- a/src/EloquentDataTable.php
+++ b/src/EloquentDataTable.php
@@ -140,7 +140,7 @@ class EloquentDataTable extends QueryDataTable
         $table     = '';
         $deletedAt = false;
         $lastQuery = $this->query;
-        foreach (explode('.', $relation) as $eachRelation) {
+        foreach (explode('.', $relation) as $index => $eachRelation) {
             $model = $lastQuery->getRelation($eachRelation);
             switch (true) {
                 case $model instanceof BelongsToMany:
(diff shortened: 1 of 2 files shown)


src/EloquentDataTable.php
```diff
@@ -169,19 +169,23 @@ protected function joinEagerLoadedColumn($relation, $relationColumn)
 
                 case $model instanceof BelongsTo:
                     $table     = $model->getRelated()->getTable();
+                    $alias     = "alias_{$index}_{$table}";
+                    $tableAs   = "{$table} AS $alias";
```
yajra: This will break some installations with Oracle database. Please use `"{$table} $alias"` removing the `AS` from query.

Morinohtar: But if i remove the `AS` then MySQL won't work...

Morinohtar: Can you run your tests again now?

yajra: Sorry but you previous PR version is much better. AFAIK, it should work on MySQL since that is one of the reasons why I created this package version to work on Oracle.

On the other hand, I think I already created a patch that properly translates Oracle `AS` query. Will review with my other package https://github.com/yajra/laravel-oci8 too.

Morinohtar: So, what now? You want the first one again?

src/EloquentDataTable.php
```diff
@@ -38,6 +45,9 @@ public function __construct($model)
         parent::__construct($builder->getQuery());
 
         $this->query = $builder;
+
+        $connection   = config('database.default');
```
yajra: There are cases where the connection is set from the model or via runtime using query. This is will not work on those situations.

Morinohtar: Ok, sure. But in your earlier reply (https://github.com/yajra/laravel-datatables/pull/2051#discussion_r274738567) you said, basically, this wouldnt even be necessary. So, what now?

yajra: Yes, so please revert the commits to 1st one you sent. Might be able to test this within the week. Thanks!

Morinohtar: Ok, done. Sorry for the mess with the commits.

yajra:
This also fixes a lot of issues: [#2045], [#2054], [#2024], [#1977], [#880], [#577], [#522], etc.

Thanks a lot 🚀 released on v9.1.0 🍻

amrmahrous:
please consider the soft delete , the new release cause an error "cloumn related_table.deleted_at not fount in where clause" however it shall be related_table_0_deleted_at

I rollback back to 9.0.1 and it works just perfect .

thank you for the great work , this project helps me a lot making fast delevlopment !

yajra:
Reverting due to unseen breaking change on soft deletes.

Related issue: https://github.com/yajra/laravel-datatables/issues/2058

thomaswalker-dev:
Any news on this?

Morinohtar:
@yajra is this issue already addressed in the latest version or is it still going on?

carlosjfernandes:
This fix not working

moeezs82:
public function query(Jobseeker $model): QueryBuilder
    {
        return $model->applyScopes($model->newQuery()->with([
            'cv:id,jobseeker_id,name',
            'jobseeker_ques_opt.options',
        ]))->latest();
    }

this is my query for datatable it is working perfect but search is not working on relational data

PR #2080, opened 2019-05-07
[9.0] Pass global search to filter callback.

apreiml:
I've columns that shouldn't be filtered on global search. E.g. I've a checkbox filter that toggles deleted entries. I tried to filter them like so:

```php
            ->filterColumn('deleted_at', function ($query, $input) {
                if ($input !== '1') {
                    $query->whereNull('users.deleted_at');
                }
            })
```

But this would kill the global search, because this query is also applied in a `or` group.

With this change I would be able to check, if the filter method is global or not like so:

```php
            ->filterColumn('deleted_at', function ($query, $input, $isGlobalSearch) {
                if ($input !== '1' && !$isGlobalSearch) {
                    $query->whereNull('users.deleted_at');
                }
            });
```

What do you say about this solution? Is there a different approach possible?

### DIFF
--- a/src/QueryDataTable.php
+++ b/src/QueryDataTable.php
@@ -286,7 +286,7 @@ class QueryDataTable extends DataTableAbstract
 
             if ($this->hasFilterColumn($column)) {
                 $keyword = $this->getColumnSearchKeyword($index, $raw = true);
-                $this->applyFilterColumn($this->getBaseQueryBuilder(), $column, $keyword);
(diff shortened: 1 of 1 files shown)


yajra:
PR looks good but for cases like this, my usual approach is to apply the query outside the dataTables. Something like below: 

```php
$query = User::query();

if ($input !== '1') {
  $query->whereNull('users.deleted_at');
}

return DataTables::of($query)->toJson();
```

PR #2084, opened 2019-05-09
[9.0] Enable the dotted notation in the ->only() function

Arkhas:
The `->only()` function is currently not removing nested Column when using it like this : 

```
$dataTable->only(‘user.posts’);
```

using the `Arr::dot()` helper and then the `Arr::set()` helper enable the use of the dotted notation in this function

### DIFF
--- a/src/Processors/DataProcessor.php
+++ b/src/Processors/DataProcessor.php
@@ -179,7 +179,17 @@ class DataProcessor
         if (is_null($this->onlyColumns)) {
             return $data;
         } else {
-            return array_intersect_key($data, array_flip(array_merge($this->onlyColumns, $this->exceptions)));
+            $dotted_array = array_intersect_key(
+                array_dot($data),
+                array_flip(array_merge($this->onlyColumns, $this->exceptions))
+            );
+
+            $nested_array = [];
+            foreach ($dotted_array as $key => $value) {
+                array_set($nested_array, $key, $value);
+            }
+
+            return $nested_array;
         }
     }
 
--- a/src/Processors/DataProcessor.php
+++ b/src/Processors/DataProcessor.php
@@ -180,8 +180,13 @@ class DataProcessor
             return $data;
         } else {
             $results = [];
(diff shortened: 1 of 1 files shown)


yajra:
Looks like it's working when picking selected relationship fields. However, it doesn't seem to work when picking only the related model?

```php
// query
$articles = Article::query()->with('category', 'creator')->select('articles.*');

// select specific fields and relation
->only(['title', 'category']) // NO category on response
```

This returns the title but the category is not displayed. 

If we specify a field on category model, it returns fine.

```php
->only(['title', 'category.id']) // WORKING
```

Arkhas:
Indeed, I adjust the code to reflect that
(and take the incoming deprecation of `array_` functions)

yajra:
@Arkhas the values becomes null when picking a specified relation column.

```php
$query = Role::with('permissions');

->only(['id', 'permissions.name']) //not working, returns null value for permissions.name

->only(['id', 'permissions']) // working
```

yajra:
UPDATE: maybe because permissions is a hasMany relationship which causes the issue. I think this would work on a belongsTo relationship.

Arkhas:
> EDIT: since permissions is an array, maybe we could use a wildcard?
> 
> ```
> ->only(['id', 'permissions.*.name']) 
> ```

the data_get function would work fine with this notation : `'permissions.*.name'`

The issue is when rebuilding the results array, you can loose the link between your relation as it is not keyBy the id of the object, maybe allowing the dotted notation for `hasMany/belongsToMany` could be another PR ?

yajra:
@Arkhas was thinking the same about a separate PR for `hasMany` relationships.

Anyways, found another issue. Why does the PR always include the `DT_*` attribute columns on response? 

image

Works well with belongsTo relationship btw.

Arkhas:
the difference between the array_intersect_key and the array_merge is that it ignore the missing columns in the data array. I fix it !

yajra:
Released on v9.4.0, thanks! 🚀

PR #2085, opened 2019-05-09
[9.0] Add the possibility to makeHidden() some attribute of a model.

Arkhas:
When converting the data into an array processable by the library, if the row object is an instance of Arrayable, the method `toArray()` is called on it.

The `toArray()` method compute all getters on the object’s attributes which can take some processing time.

Within this PR, I added the possiblity to use a makeHidden() before this convertion using the following synthax : 

```
$datatable->makeHidden([’posts`]);
```

It can grealty reduce the processing time if you don’t need a huge relationship in the resulting array

As for now, the `makeHidden()` of an eloquent Object does not allow the use of a dot notation (like this `->makeHidden([’posts.comments`]);`) but I hope it will eventually. In the meantime, it might be usefull to implement it inside the laravel-datatables library.

### DIFF
--- a/src/Utilities/Helper.php
+++ b/src/Utilities/Helper.php
@@ -151,10 +151,12 @@ class Helper
      * Converts array object values to associative array.
      *
      * @param mixed $row
+     * @param array $filters
      * @return array
      */
-    public static function convertToArray($row)
+    public static function convertToArray($row, $filters = [])
     {
+        $row = method_exists($row, 'makeHidden') ? $row->makeHidden($filters['make_hidden']) : $row;
         $data = $row instanceof Arrayable ? $row->toArray() : (array) $row;
 
         foreach ($data as &$value) {
--- a/src/Utilities/Helper.php
+++ b/src/Utilities/Helper.php
@@ -120,7 +120,7 @@ class Helper
     public static function castToArray($param)
     {
         if ($param instanceof \stdClass) {
-            $param = (array)$param;
+            $param = (array) $param;
 
             return $param;
         }
@@ -140,7 +140,7 @@ class Helper
      */
     public static function getOrMethod($method)
     {
-        if (!Str::contains(Str::lower($method), 'or')) {
+        if (! Str::contains(Str::lower($method), 'or')) {
             return 'or' . ucfirst($method);
         }
 
@@ -157,7 +157,7 @@ class Helper
     public static function convertToArray($row, $filters = [])
     {
         $row  = method_exists($row, 'makeHidden') ? $row->makeHidden($filters['make_hidden']) : $row;
-        $data = $row instanceof Arrayable ? $row->toArray() : (array)$row;
+        $data = $row instanceof Arrayable ? $row->toArray() : (array) $row;
 
         foreach ($data as &$value) {
             if (is_object($value) || is_array($value)) {
@@ -194,7 +194,7 @@ class Helper
                 $row[$key] = $value->format('Y-m-d H:i:s');
             } else {
                 if (is_object($value)) {
-                    $row[$key] = (string)$value;
+                    $row[$key] = (string) $value;
                 } else {
                     $row[$key] = $value;
                 }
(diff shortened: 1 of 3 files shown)


FabienLucini:
👍

Arkhas:
THe big difference between the removecolumn and a `makeHidden()` is that the `makeHidden` is made BEFORE the `toArray()` on the object and the `removeColumn()` is done once all the processing is done on the computed array() and it only remove data from it while the `makeHidden()` prevent the attribute to be process at all.

as an example : 

Your query is like this : 

```
$query = User::with('posts.comments');
```

You have these columns : 

| ID          | User Name     | Last Post                                | Last comment's user on Last Post                                 | Average Rating                |
|-------------|---------------|------------------------------------------|------------------------------------------------------------------|-------------------------------|
| `$user->id` | `$user->name` | `$user->posts->latest()->first()->title` | `$user->posts->latest()->first()->comments->first()->user->name` | `$user->posts->avg('rating')` |
|             |               |                                          |                                                                  |                               |
|             |               |                                          |                                                                  |                               |

It's just as an example, all of theses could be done on raw SQL, but bear with me

You dont want neither of your posts and comments in your JSON so with remove column, you would do this: 

```
$datatable->removeColumn('posts');
```

BUT, the `$row->toArray()`  is still processing all of the User Posts AND each Posts's Comments to store it inside an array THEN it is unset

While doint this : 

```
$datatable->makeHidden('posts');
```

when doing `$row->toArray()`, all the `Posts` are skipped from being add to the array and neither are the comments

with #2084, the `removeColumn()` method allow the use of the dotted notation, so if you want your posts in your JSON but not the comments, you can do this : 

```
$datatable->removeColumn('posts.comments');
```

but you cant' do this 

```
$datatable->makeHidden('posts.comments');
```

because the `toArray()` function is not working with a dotted notation.

HadrienKulik:
On tables of thousands of lines, for which we have to load deep relations to get some attributes to display in the columns, this `toArray()` on the giant collection at the very beginning of the processing is clearly killing the performances. So yes, having a way of clearing the useless beforehand would be useful I think, especially when we use the "export" functions (excel, csv,..) which need to process the whole table and not just the displayed page.

On top of that, the `toArray()` triggers the computation of all attributes, even attributes that can be complex with processing associated to them (`getFooAttribute()`) even though these attributes are not necessarily displayed in the table in the end.

Wouldn't be a way to remove this `toArray()` ideally ? Not sure I anticipate all the use cases, but in ours, when using Eloquent, it seems to be a performance bottleneck.

Btw, in serverside mode, one can wonder if the use of the only() method to clean the returned json by only returning the necessary data that are going to be displayed in the columns should not be the default behavior also. We do it by hand by taking our list of columns and giving it to the only() method, because there is not really any reason to return more than those columns, but I feel like in serverside mode that should be done by default for the user.

Arkhas:
@yajra Do you see anything else preventing to implementing this in the library ?

yajra:
@Arkhas got a bit occupied lately. Checking this again. Thanks!

src/Utilities/Helper.php
```diff
@@ -151,10 +151,12 @@ public static function getOrMethod($method)
      * Converts array object values to associative array.
      *
      * @param mixed $row
+     * @param array $filters
      * @return array
      */
-    public static function convertToArray($row)
+    public static function convertToArray($row, $filters = [])
     {
+        $row  = method_exists($row, 'makeHidden') ? $row->makeHidden(array_get($filters, 'hidden', [])) : $row;
```
yajra: Maybe we should avoid array functions due to deprecation like on your other PR? 

Anyways, I think there are other array functions in the package that needs to be updated maybe on another PR. A minor issue and this is still good for now.

yajra:
Just to be sure, this is not intended to hide a specific column on relations right? Something like:

```php
$query = Role::with('permissions');

->makeHidden(['permissions.name']) //not working

->makeHidden(['permissions']) // working
```

yajra:
After reviewing further, I realised that we are only piping to Eloquent collection function thus this would only work on how collection makeHidden function works.

yajra:
Released on v9.3.0 🚀 ... If possible, can you please submit a PR for [docs](https://github.com/yajra/laravel-datatables-docs)? Thanks! 🍻

Issue #1793, opened 2018-07-07
Malformed UTF-8 characters, possibly incorrectly encoded, when use uuid field 


bpartner:
### Summary of problem or feature request

Log:
InvalidArgumentException: Malformed UTF-8 characters, possibly incorrectly encoded in D:\OpenServer\domains\finkpi.local\vendor\laravel\framework\src\Illuminate\Http\JsonResponse.php:75 Stack trace: #0 D:\OpenServer\domains\finkpi.local\vendor\symfony\http-foundation\JsonResponse.php(50): Illuminate\Http\JsonResponse->setData(Array) #1 
..........

Ajax response
{"draw":1,"recordsTotal":1,"recordsFiltered":0,"data":[],"error":"Exception Message:\n\nMalformed UTF-8 characters, possibly incorrectly encoded"}
------------------------
id & user_id is BINARY, 

"id" => b"\x11Þü╩?¾$Éº[\x00 \x05(!>"
"user_id" => b"\x11Þü┼9¿ÃÈ¼\x01\x00 \x05(!>"

### Code snippet of problem

        $auth_user_id = Auth::user()->uuid_text;
        $models = Motivation::withUuid($auth_user_id,'user_id');
        $datatables = DataTables::eloquent($models)
                         ->setTransformer(new MotivationTransformer)
                         ->make();

Transformer

        return [
            'id' => $models->uuid_text,
            'name' => $models->name,
            'active' => $models->active ? '<span class="label bg-success-300" data-i18n="page.models.mm_table.active">Active</span>':'<span class="label bg-danger-300" data-i18n="page.models.mm_table.pending">Pending</span>',
            'created_at' => Carbon::parse($models->created_at)->format('d-m-Y'),
            'actions' =>
                '<button class="edit btn text-info-600 btn-flat btn-icon" data-id="'.$models->uuid_text.'">
                       <i class="icon-database-edit2"></i></button>
                 <button class="delete btn text-warning-600 btn-flat btn-icon" data-id=" '.$models->uuid_text.'">
                       <i class="icon-database-remove"></i></button>'
        ];

### System details

- Windows 10
- PHP 7.2
- Laravel 5.6
- Laravel-Datatables 8

Used https://github.com/spatie/laravel-binary-uuid

bpartner:
Problem in method render in abstract class DataTableAbstract,
when the debugging mode is on 

protected function render(array $data)
    {
        $output = $this->attachAppends([
[...]

PR #2088, opened 2019-05-13
[9.0] Prevent malformed UTF-8 characters in debug mode.

drsdre:
To prevent UTF-8 error messages in debug mode when queries contain binary data, all texts in the queries part are processed with utf8_encode before adding in to the output.

### DIFF
--- a/src/QueryDataTable.php
+++ b/src/QueryDataTable.php
@@ -724,7 +724,13 @@ class QueryDataTable extends DataTableAbstract
      */
     protected function showDebugger(array $output)
     {
-        $output['queries'] = $this->connection->getQueryLog();
+	    $query_log = $this->connection->getQueryLog();
+	    array_walk_recursive($query_log, function (&$item, $key)
+	    {
+		    $item = utf8_encode($item);
+	    });
+
+        $output['queries'] = $query_log;
         $output['input']   = $this->request->all();
 
         return $output;


yajra:
Released on v9.3.0 🚀 Thanks a lot! 🍻

matthijs:
I am sorry to say but this is not a good solution to the bug described in #1793.

The utf8_encode function only converts data encoded with ISO-8859-1 to data encoded in UTF-8. I already described in #1793 that I'm bitten by this because the query log can contain objects.

The proper fix would be to exclude the binary data or to provide a function which is able to convert the binary data into a readable form. In #1793 it is specific to UUID data, but its possible to have all kinds of binary data.

Could you please revert this or not call utf8_encode if $item is an object?

drsdre:
@matthijs can you please check if your issue is fixed with the commit https://github.com/drsdre/laravel-datatables/commit/ef655b66aa062f5439c093b46178aa51c9c8dcba ?

matthijs:
@drsdre I got the same stack trace, printed below.

But this won't fix the underlying problem. You cannot assume that the data is encoded in ISO-8859-1. Say its encoded in UTF-8, you will get garbage when calling utf8_encode. The following code snippet will demonstrate the problem when using utf8_encode.

```php
$data = 'some utf8 characters: é ü € ö';
echo $data. "\n";
echo utf8_encode($data). "\n";

// Output:
// some utf8 characters: é ü € ö
// some utf8 characters: Ã© Ã¼ â¬ Ã¶
```

There is also no way to 'detect' the input encoding, so you need to know that before hand.

The original bug is all about binary data. And to display binary data you need something that can read the binary data and make something useful of it.

I do appreciate your try to fix it!

Here is the stacktrace:
```php
[2019-06-10 17:34:26] testing.ERROR: Cannot use object of type DateTime as array {"userId":"43f82cf4-5025-44ef-b7b1-e79dcb99e3fa","exception":"[object] (Symfony\\Component\\Debug\\Exception\\FatalThrowableError(code: 0): Cannot use object of type DateTime as array at /home/matthijs/Projects/frontend/vendor/yajra/laravel-datatables-oracle/src/QueryDataTable.php:749)
[stacktrace]
#0 /home/matthijs/Projects/frontend/vendor/yajra/laravel-datatables-oracle/src/QueryDataTable.php(751): Yajra\\DataTables\\QueryDataTable->recursiveLogEncode(Object(DateTime))
#1 /home/matthijs/Projects/frontend/vendor/yajra/laravel-datatables-oracle/src/QueryDataTable.php(751): Yajra\\DataTables\\QueryDataTable->recursiveLogEncode(Array)
#2 /home/matthijs/Projects/frontend/vendor/yajra/laravel-datatables-oracle/src/QueryDataTable.php(751): Yajra\\DataTables\\QueryDataTable->recursiveLogEncode(Array)
#3 /home/matthijs/Projects/frontend/vendor/yajra/laravel-datatables-oracle/src/QueryDataTable.php(733): Yajra\\DataTables\\QueryDataTable->recursiveLogEncode(Array)
#4 /home/matthijs/Projects/frontend/vendor/yajra/laravel-datatables-oracle/src/DataTableAbstract.php(744): Yajra\\DataTables\\QueryDataTable->showDebugger(Array)
#5 /home/matthijs/Projects/frontend/vendor/yajra/laravel-datatables-oracle/src/QueryDataTable.php(104): Yajra\\DataTables\\DataTableAbstract->render(Array)
#6 /home/matthijs/Projects/frontend/app/Http/Controllers/InvestmentportfolioController.php(102): Yajra\\DataTables\\QueryDataTable->make()
#7 [internal function]: App\\Http\\Controllers\\InvestmentportfolioController->index(Object(Illuminate\\Http\\Request))
#8 /home/matthijs/Projects/frontend/vendor/laravel/framework/src/Illuminate/Routing/Controller.php(54): call_user_func_array(Array, Array)
#9 /home/matthijs/Projects/frontend/vendor/laravel/framework/src/Illuminate/Routing/ControllerDispatcher.php(45): Illuminate\\Routing\\Controller->callAction('index', Array)
#10 /home/matthijs/Projects/frontend/vendor/laravel/framework/src/Illuminate/Routing/Route.php(219): Illuminate\\Routing\\ControllerDispatcher->dispatch(Object(Illuminate\\Routing\\Route), Object(App\\Http\\Controllers\\InvestmentportfolioController), 'index')
```

yajra:
Maybe we can just omit / replace the object with something else that describes the value?

drsdre:
@matthijs going back to the original idea that this is debug information which is for human consumption only. As an alternative, I propose to go back to standard print_r PHP functionality to flatten the query data and run it through a regex filter which removes all non-printable characters. This simplies the showDebugger function to:

	protected function showDebugger( array $output ) {
		// Flatten 
		$output['queries'] = preg_replace(
			'/[[:^print:]]/',
			'',
			print_r( $this->connection->getQueryLog(), true )
		);
		
		$output['input']   = $this->request->all();

		return $output;
	}

Can you test if this works in your situation too?

matthijs:
I'll test it when back at home.

matthijs:
@drsdre I've tested this function and that works.

I also checked it in the browser and that looks as a workable solution 👍

PR #2103, opened 2019-06-03
[8.0] Removal of redundant SoftDelete check.

selecod:
Existing method of SoftDelete check brakes relations that looks like:
```
class Thread extends Model {
   public function threads() {
      return $this->belongsTo(Thread::class)->withTrashed();
   }
}
```

or

```
class Thread extends Model {
   public function threads() {
      return $this->belongsTo(Thread::class)->onlyTrashed();
   }
}
```
Both of them got filtered by `whereNull($deletedAt)` while joining. 
Instead of this, no filtering should be applied at all on first model. 
`whereNotNull($deletedAt)` should be called at second model. 
This works well out of a box.

### DIFF
--- a/src/EloquentDataTable.php
+++ b/src/EloquentDataTable.php
@@ -138,7 +138,6 @@ class EloquentDataTable extends QueryDataTable
     protected function joinEagerLoadedColumn($relation, $relationColumn)
     {
         $table     = '';
-        $deletedAt = false;
         $lastQuery = $this->query;
         foreach (explode('.', $relation) as $eachRelation) {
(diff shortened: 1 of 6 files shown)


yajra:
Need to review this further. Is it possible to have some tests for this? Will also to write some more tests. Thanks!

selecod:
I've added some test for `HasOne` and `HasMany` relations, hope it'll help.

yajra:
Released on v8.13.7 and v9.4.1 🚀 . Thanks a lot!

PR #2183, opened 2019-09-20
[7.0] Use an empty array as the default value when no columns are found

u01jmg3:
At my place of work we are still using Laravel 5.4 and Laravel Datatables 7. This small change is necessary to prevent us requiring a fork of Laravel Datatables.

Here are the later commits where this problem is fixed in Laravel Datatables 8/9.

- https://github.com/yajra/laravel-datatables/commit/4136d7b
- https://github.com/yajra/laravel-datatables/commit/784fdeb


yajra:
Released on v7.10.3, thanks! 🚀

PR #2193, opened 2019-09-28
[9.0] Add column formatter.

yajra:
- Proposed implementation of field/column [formatters](https://editor.datatables.net/manual/php/formatters).
- Fix https://github.com/yajra/laravel-datatables-editor/issues/33

### Date Formatter

```php
use Carbon\Carbon;
use Yajra\DataTables\Contracts\Formatter;

class DateFormatter implements Formatter
{
    protected $format;

    public function __construct($format = null)
    {
        $this->format = $format;
    }

    public function format($value, $row)
    {
        if ($this->format) {
            return Carbon::parse($value)->format($this->format);
        }

        return Carbon::parse($value)->diffForHumans();
    }
}
```

### Server-side script
```php
    public function dataTable($query)
    {
        $formatter = new DateFormatter($format = 'Y-m-d');

        return datatables()
            ->eloquent($query)
            ->formatColumn('updated_at', new DateFormatter) // using new formatter instance
            ->formatColumn('created_at', DateFormatter::class) // using formatter class
            ->formatColumn('published_at', $formatter) // using instance
            ->formatColumn(['published_at', 'created_at'], $formatter) // format multiple columns
            ->setRowId('id');
    }

 ...

// columns
return [
  Column::make('status')->data('status_formatted'),
  ...
]
```

### Json Response

```js
{
  data: [
    {
      id: 1, 
      created_at: "2020-11-03T02:27:20.000000Z", 
      created_at_formatted: "2020-11-03"
    }
  ],
  ...
}
```

### DIFF
--- a/src/Contracts/Formatter.php
+++ b/src/Contracts/Formatter.php
@@ -1,0 +1,13 @@
+<?php
+
+namespace Yajra\DataTables\Contracts;
+
+interface Formatter
+{
+    /**
+     * @param string $value
+     * @param mixed $row
+     * @return string
+     */
+    public function format($value, $row);
+}
--- a/src/DataTableAbstract.php
+++ b/src/DataTableAbstract.php
@@ -2,17 +2,18 @@
 
 namespace Yajra\DataTables;
 
-use Illuminate\Contracts\Support\Arrayable;
-use Illuminate\Contracts\Support\Jsonable;
-use Illuminate\Http\JsonResponse;
 use Illuminate\Support\Arr;
 use Illuminate\Support\Str;
-use Illuminate\Support\Traits\Macroable;
 use Psr\Log\LoggerInterface;
-use Yajra\DataTables\Contracts\DataTable;
-use Yajra\DataTables\Exceptions\Exception;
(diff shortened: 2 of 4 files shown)


OzanKurt:
Just keep in mind that creating hundreds of `View` instances is not good for the application. It's usually a lot better to generate the html "UGLY but inline" inside of the class directly.

OzanKurt:
Hey @yajra ,

I was recently doing some KPI analytics and sadly encountered the same problem as this dev: https://github.com/yajra/laravel-datatables/issues/3016

The thing is;
- if I get `$query->first()->avg_profile_quality` its `2.3`
- if I get `$query->toArray()` the `avg_profile_quality` is still `2.3`
- and yet also in database its `2.3`

image

But for some reason when it's generating the response, all the data is converted to something like this:

image

Could you please have a look at this?

OzanKurt:
After further research I found out that it's a PHP/Laravel problem.

Setting the `serialize_precision` config of php inside my `index.php` solved the problem.

```
ini_set('serialize_precision', -1);
```

It's still extremely weird how such thing is possible.

PR #2199, opened 2019-10-02
[9.0] Allow orderColumn using closure.

### LINKED ISSUE #2159: OrderCoumn with Closure

We can only use raw sql in the orderColumn function, it would be nice if we would be able to something like : 

```
->orderColumn('name', function($query, $order){
      return $query->orderBy('name', $order);
});
```

that would allow more complexe ordering like : 

```
->orderColumn('user.name', function($query, $order){
      return $query->join('users', 'user_id', '=', 'users.id')
                             ->orderBy('users.name', $order);
});
```

yajra:
Fix #2159, #2194.

```php
return datatables()
    ->eloquent($query)
    ->setRowId('id')
    ->orderColumn('status', function($query, $order) {
        $query->orderBy('status', $order);
    });
```

### DIFF
--- a/src/QueryDataTable.php
+++ b/src/QueryDataTable.php
@@ -673,10 +673,14 @@ class QueryDataTable extends DataTableAbstract
      */
     protected function applyOrderColumn($column, $orderable)
     {
-        $sql      = $this->columnDef['order'][$column]['sql'];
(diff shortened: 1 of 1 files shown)


Arkhas:
Sounds good to me ! Thanks !

PR #2261, opened 2019-12-06
Added multiple search

dima-bzz:
Added multiple search for CollectionDataTable. If multiple => true, multi_term disable.

### DIFF
--- a/src/CollectionDataTable.php
+++ b/src/CollectionDataTable.php
@@ -21,6 +21,13 @@ class CollectionDataTable extends DataTableAbstract
      *
      * @var \Illuminate\Support\Collection
      */
+    public $merged;
+
+    /**
+     * Collection object.
+     *
+     * @var \Illuminate\Support\Collection
+     */
     public $original;
 
     /**
@@ -66,6 +73,7 @@ class CollectionDataTable extends DataTableAbstract
         $this->request    = app('datatables.request');
         $this->config     = app('datatables.config');
         $this->collection = $collection;
+        $this->merged     = collect();
         $this->original   = $collection;
         $this->columns    = array_keys($this->serialize($collection->first()));
     }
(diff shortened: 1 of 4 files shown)


yajra:
Sorry for late response. Will try to review this as soon as I can. Thanks!

yajra:
@dima-bzz can you please elaborate the difference of this PR from `multi_term` search? Can you please provide some use case? Thanks!

src/CollectionDataTable.php
```diff
@@ -250,6 +258,49 @@ protected function globalSearch($keyword)
         });
     }
 
+    /**
+     * Perform multiple search for the given keyword.
+     *
+     * @param string $keyword
+     */
+    protected function multiSearch($keywords)
```
yajra: This seems like a duplicate of `globalSearch` method.

src/config/datatables.php
```diff
@@ -16,6 +16,11 @@
          */
         'multi_term'       => true,
 
+        /*
+         * Multiple search will explode search keyword using spaces resulting into multiple search.
+         */
+        'multiple'       => false,
```
yajra: Multi_term already explodes the keyword using spaces.

```php
    /**
     * Perform multi-term search by splitting keyword into
     * individual words and searches for each of them.
     *
     * @param string $keyword
     */
    protected function smartGlobalSearch($keyword)
    {
        collect(explode(' ', $keyword))
            ->reject(function ($keyword) {
                return trim($keyword) === '';
            })
            ->each(function ($keyword) {
                $this->globalSearch($keyword);
            });
    }
```

dima-bzz:
@yajra Mutli_term searches through a space, but it works through a loop and each next iteration is searched in the previous one.  Multiple search searches for everything in a single query, without looping. For example, I have a list of users and I need to find all users with Administrator and User rights. Through multi_term there will be only the Administrator, and through multiple search both the Administrator and the User.

yajra:
@dima-bzz thanks for elaborating. What you described should be the functionality of multi-term and thus it might be a bug on the current implementation. Maybe we should apply your PR to multi-term instead of creating a new multiple config?

dima-bzz:
@yajra possible. But on the other hand, multi_term also has its own plus, it is a sequential search. For example, I first found all users with Administrator rights, and then found a specific user. Just for different tasks is convenient or multi_term or multiple search. You can simply combine the duplicated code, but leave both methods.

dima-bzz:
@yajra removed duplicate code. Left both search methods.

dima-bzz:
@yajra What about my PR?

yajra:
@dima-bzz a bit busy atm and can't find time to review this yet. Will try to give this some time within the week. Or if anyone in the community can review this would be a great help.

PR #2274, opened 2020-01-06
Adds Option to Enable/Disable Searching & Filtering (In case user already has searched/filtered dataset)

dylancoots-mwi:
This pull request contains changes that would allow the developer to choose if they would like to use the DataTables built in Search and Filtering. Over time using this package, I found that it's best to handle the searching and filtering before passing the data into DataTables. All default to current default activity (search = true, sort = true)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,5 +1,5 @@
 {
-  "name": "dylancoots-mwi/laravel-datatables",
+  "name": "yajra/laravel-datatables-oracle",
   "description": "jQuery DataTables API for Laravel 4|5",
   "keywords": [
     "laravel",
--- a/composer.json
+++ b/composer.json
@@ -1,5 +1,5 @@
 {
-  "name": "dylancoots-mwi/laravel-datatables-oracle",
+  "name": "yajra/laravel-datatables-oracle",
   "description": "jQuery DataTables API for Laravel 4|5",
   "keywords": [
     "laravel",
--- a/composer.json
+++ b/composer.json
@@ -1,5 +1,5 @@
 {
-  "name": "yajra/laravel-datatables-oracle",
+  "name": "dylancoots-mwi/laravel-datatables",
   "description": "jQuery DataTables API for Laravel 4|5",
   "keywords": [
     "laravel",
--- a/composer.json
+++ b/composer.json
@@ -1,5 +1,5 @@
 {
-  "name": "dylancoots-mwi/laravel-datatables",
+  "name": "yajra/laravel-datatables-oracle",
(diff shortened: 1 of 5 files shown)


yajra:
Thanks but I think you can already disable searching / ordering on the javascript side? Just set the ff options:

```js
$('table').DataTable({
    searching: false,
    ordering: false,
    ...
});
```

dylancoots-mwi:
> Thanks but I think you can already disable searching / ordering on the javascript side? Just set the ff options:
> 
> ```js
> $('table').DataTable({
>     searching: false,
>     ordering: false,
>     ...
> });
> ```

Yes but that also disables the search box entirely and this pull request would allow developers to handle search their own way while keeping other functionality intact. Same with sorting.

yajra:
I see, another work around is to use an empty closure. Do you think the code below would work for you? I wanted to avoid using `make()` and promote `toJson()` instead if possible.

```php
        return datatables()
            ->eloquent($query)
            ->order(function () {})
            ->filter(function () {})
...
```

dylancoots-mwi:
May I ask why you prefer to promote toJson()?

This is my use-case, I just find it cleaner to not be required to pass empty closures along through a chain when the data passed in is already finished being manipulated.




```
    if (request()->has('dt')) {
       // Everything I want handled is done here outside of DataTables
       // This method uses another Package that handles eloquent text search using joins, when DataTables does not - this is the key reason for the PR, it allows use of other search packages by bypassing DataTables search.
        $jobs = JobService::searchJobs(
            request()->search['value'] ?? null,
            request()->length ?? null,
        );

        return DataTables::of($jobs)
            ->setTransformer(new JobTransformer)
            ->make(true, false, false); // If PR is merge, instead of your suggestion.
    }```

yajra:
> May I ask why you prefer to promote toJson()?

I wanted to drop support for `make(false)` and `toJson() == make(true)` :).

What about updating the methods above that would disable searching and sorting which basically wraps the code above? Maybe something like:

```php
        return DataTables::of($jobs)
            ->order(false)
            ->filter(false)
            ->toJson();

dylancoots-mwi:
> > May I ask why you prefer to promote toJson()?
> 
> I wanted to drop support for `make(false)` and `toJson() == make(true)` :).
> 
> What about updating the methods above that would disable searching and sorting which basically wraps the code above? Maybe something like:
> 
> ```
>         return DataTables::of($jobs)
>             ->order(false)
>             ->filter(false)
>             ->toJson();
> ```

I think that would be better than forcing a closure. Thanks Yajra!  Your solution with closures works great; however, i recommend updating to allow `false` as a param!

PR #2281, opened 2020-01-13
Fix TotalCount logic

romanstingler:


yajra:
Can you please provide some snippets to reproduce the issue that is being fixed here? Thanks!

romanstingler:
@yajra hey sorry, I was totally wrong with this commit. 
I have looked at the code and will provide a real solution this weekend.

karmendra:
I do not understand the fix here. What exactly is the issue with TotalCount logic?

PR #2380, opened 2020-04-23
Added orthogonal sort and filter for collection

dima-bzz:
Added support orthogonal. So if you sort, the name is used from data.sort. If you search, the name is used from data.filter.

**PHP**
```php
...
->editColumn('updated_at', function ($row) {
                return [
                    'display' => $row->getUpdatedAt(),
                    'timestamp' => $row->getUpdatedAtTimestamp(),
                ];
            })
...
```

**JS**
```js
...
{'name': 'updated_at', 'data': { '_': 'updated_at.display', 'sort': 'updated_at.timestamp', 'filter': 'updated_at.display'}},
...
```


yajra:
Can you please provide some refs on the official js docs? Does the orthogonal data also being passed when using server-side? Not sure but I think this only works on client-side.

dima-bzz:
@yajra https://datatables.net/manual/server-side#Returned-data


https://datatables.net/reference/option/columns.data

yajra:
Thanks for the refs. I think this still not supported on serverSide option. The [ajax example](https://datatables.net/examples/ajax/orthogonal-data) loads all data at once and `serverSide: false` is the default settings thus this is only usable on clientSide implementation.

Also, when you try to inspect the ajax request on serverSide request. `display` and `_` are not sent by default. Attaching the columns data being sent.

image

You have a working unit test, but it's not how the js library works.

dima-bzz:
@yajra I do not quite understand you, where does ajax and `serverSide: false`. On the server, I edit the updated_at column:
```php
...
->editColumn('updated_at', function ($row) {
                return [
                    'display' => $row->getUpdatedAt(),
                    'timestamp' => $row->getUpdatedAtTimestamp(),
                ];
            })
...
```
Data comes to the client:
```
{
    "draw": 2,
    "recordsTotal": 32,
    "recordsFiltered": 32,
    "data": [
        {
            "type": "Mail",
            "updated_at": {
                "display": "27 апреля 2020 г., 15:40",
                "timestamp": "1587991247"
            },
...
```

In JS code, I specify column and `serverSide: true`:
```js
...
{'name': 'updated_at', 'data': { '_': 'updated_at.display', 'sort': 'updated_at.timestamp', 'filter': 'updated_at.display'}},
...
```

And when I sort the updated_at column or search globally, the data is sent to the server like this:
```
..
columns[2][data][_]: updated_at.display
columns[2][data][sort]: updated_at.timestamp
columns[2][data][filter]: updated_at.display
columns[2][name]: updated_at
columns[2][searchable]: true
columns[2][orderable]: true
columns[2][search][value]
columns[2][search][regex]: false
...
```

Accordingly, I want the server to search by name from columns [2] [data] [filter]: updated_at.display and sort from columns [2] [data] [sort]: updated_at.timestamp. Which I implemented for the Collection engine

yajra:
My bad, it seems like the full column is being sent as I forgot to set the columns data to orthogonal. Will review this again.

yajra:
The PR looks good. 

Would you be able to create a test on https://github.com/yajra/laravel-datatables/blob/9.0/tests/Integration/CollectionDataTableTest.php that emulates this feature please?

PR #2385, opened 2020-04-30
Throw the original exception

mikepsinn:
Throw the original exception so you can find out where it originated.   Currently, it's a pain to track down and fix the source of exceptions even if 'throw' is set, because they only include the message.


yajra:
This would break current installations that catches the custom exception. I think the full error would still be on laravel log file and you can track it from there?

mikepsinn:
Got it. Thanks, @yajra!🙂

PR #2391, opened 2020-05-12
[9.0] Fix issue with non-existing variables/attributes .

mgralikowski:
This change fix problem with non-existing variables/attributes during sorting by Collection engine.

There is no problem with displaying undefined values thanks this:

`->columnDefs([['defaultContent' => '-', 'targets' => '_all']])`

But when we try sort column we will get an error:

`Undefined index: resourcename.non_existing_attribute in C:\laragon\www\app\vendor\yajra\laravel-datatables-oracle\src\CollectionDataTable.php:298`

Screenshot_1


yajra:
Released on v9.10.1, thanks!

Issue #1975, opened 2019-01-30
Response is casting all attributes to string


D4ni3lr:
**Al attributes are being casts as String**

All attributes are being casts  to String, even when the `$casts` attribute is present in the model.

I've tried with

```
Datatables::of(Model::get())->toJson();
Datatables::of(Model::get())->make();
datatables()->of(Model::get())->toJson();
datatables()->of(Model::get())->make();
```
Same result.

When i return `Model::get();`  or  `Model::all();` the attribute types are returned as expected.

**System Details**

- Windows 10
- Apache 2.4
- PHP 7.1
- laravel/framework  v5.5.45
- yajra/laravel-datatables              v1.4.0  Laravel DataTables Complete Package.
- yajra/laravel-datatables-buttons      v4.5.0  Laravel DataTables Buttons Plugin.
- yajra/laravel-datatables-editor       v1.6.1  Laravel DataTables Editor plugin for Laravel 5.5+.
- yajra/laravel-datatables-fractal      v1.3.0  Laravel DataTables Fractal Plugin.
- yajra/laravel-datatables-html         v4.3.1  Laravel DataTables HTML builder plugin for Laravel 5.4+.
- yajra/laravel-datatables-oracle       v8.13.4 jQuery DataTables API for Laravel 4|5

D4ni3lr:
Hi, the problem was introduced in this PR: #1952 in file DataProcessor.php inside the function escapeRow two lines where removed. 

```
// File: src/Processors/DataProcessor.php
...
if (! in_array($key, $this->rawColumns)) {
       $dataType       = gettype($value); 
       $arrayDot[$key] = e($value);
       settype($arrayDot[$key], $dataType);
 }
...
```

Thijmen:
Really waiting for this one, any updates?

artemsky:
any update?

bendt-indonesia:
Same problem here, using PHP 7.2, Laravel 5.8, yajra/laravel-datatables-oracle 9.4.0

bendt-indonesia:
> Hi, the problem was introduced in this PR: #1952 in file DataProcessor.php inside the function escapeRow two lines where removed.
> 
> ```
> // File: src/Processors/DataProcessor.php
> ...
> if (! in_array($key, $this->rawColumns)) {
>        $dataType       = gettype($value); 
>        $arrayDot[$key] = e($value);
>        settype($arrayDot[$key], $dataType);
>  }
> ...
> ```

[...]

Issue #1747, opened 2018-05-23
Eloquent DataTables casts all attributes to string


peyobr:
### Eloquent DataTables casts all attributes to string

Eloquent DataTables casts all attributes to string, even int ids, even if explicit eloquent model $casts is set.

### System details

- Linux Mint 18.3
- PHP 7.0.30-0ubuntu0.16.04.1
- Laravel Version 5.5.40
- Laravel-Datatables Version 8.6.0
- Mysql 5.5.72-ubuntu0.16.04.1

BARNZ:
I encountered this problem today as well. Wondering why this would be happening?

What I'm getting:

![image](https://user-images.githubusercontent.com/2202345/45399988-5f043780-b69e-11e8-9a23-588df7775455.png)

What I expect:

![image](https://user-images.githubusercontent.com/2202345/45399969-4eec5800-b69e-11e8-8145-550b4638dd2e.png)

yajra:
Found the related issue/pr: https://github.com/yajra/laravel-datatables-buttons/pull/67.

ATM, I think you need to use fractal integration for proper API response but I also agree that we can improve the package not to casts all data to string and let the eloquent casting handle it. 

Can't check this further though but will definitely implement this when I got the chance. If any of you can, please do not hesitate to submit a PR. Thanks!

ridaamirini:
@yajra I added a PR

yajra:
Fixed on v8.13.2, thanks to @ridaamirini!

jovpet:
What have happened with that? It seems it has been reverted in version 9! `DataProcessor.php`
It seems this PR from @ridaamirini has been discarded. @yajra ?
`
protected function escapeColumns(array $output)
    {
        return array_map(function ($row) {
            if ($this->escapeColumns == '*') {
                $row = $this->escapeRow($row);
            } elseif (is_array($this->escapeColumns)) {
                $columns = array_diff($this->escapeColumns, $this->rawColumns);
                foreach ($columns as $key) {
                    Arr::set($row, $key, e(Arr::get($row, $key)));
                }
            }

            return $row;
        }, $output);
    }
`

yajra:
It was reverted due to a breaking change and still unresolved. https://github.com/yajra/laravel-datatables/issues/1951. Will try to have a look at this again when I got the chance.

rilwanmsm:
Any updates?

fayaz668:
[...]

PR #2399, opened 2020-05-21
Fix type casting issue

ehsanrasta:
due to #1975 , #1747 
Instead of escaping all values of row we should escape only strings.


yajra:
This looks like a better solution. Feedbacks & review would be appreciated from people on #1975 & #1747.

yajra:
Released v9.10.1 thanks!

PR #2414, opened 2020-06-09
Escape row in data processor also accepts Htmlable 

edwwaarrdd:
After the PR #2399 some of our Datatables were not working as expected.
This is because as a row attribute we sometimes pass a `Htmlable` object `HtmlString`.

Like this:
```
    public function dataTable($query): DataTableAbstract
    {
        return datatables()
            ->eloquent($query)
            ->setRowClass('cursor-pointer')
            ->addRowAttr('onclick', fn (Model $model) => new HtmlString("window.location.href = '" . route('blabla') . "';"));
    }
```
I just added an extra check to the dataprocessor function escapeRow to see if it is a Htmlable object. So it can be passed to the `e` function where it correctly handles the Htmlable
```
if ($value instanceof Htmlable) {
            return $value->toHtml();
        }
```

### DIFF
--- a/src/Processors/DataProcessor.php
+++ b/src/Processors/DataProcessor.php
@@ -2,6 +2,7 @@
 
 namespace Yajra\DataTables\Processors;
 
+use Illuminate\Contracts\Support\Htmlable;
 use Illuminate\Support\Arr;
 use Yajra\DataTables\Utilities\Helper;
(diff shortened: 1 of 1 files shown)


yajra:
Released on v9.10.2 🚀 Thanks!

PR #2478, opened 2020-10-05
Use simple numeric sorting when dealing with numeric values

### LINKED ISSUE #1934: Negative number sorting issue with collection datasource
## Improper sorting is obtained when sorting columns containing signed numeric values.
When using collection as datasource, and sorting one or more column containing signed numeric values, incorrect sort is obtained for negative values. Consider the following case for example:

**Column Input values:** `[0,200,-228,1000,500,-1000]`
**Current output on ascending sort:** `[-228,-1000,0,220,500,1000]`
**Expected output on ascending sort:** `[-1000,-228,0,220,500,1000]`

When did a digging to the issue, its found that in `CollectionDataTable@getSorter`, the functions `strnatcasecmp()` & `strnatcmp()` are being used to compare strings which considers `-228 < -1000` (-228 smaller than -1000)

### System details
- Operating System: Windows 10/Linux
- PHP Version: 7.2.3
- Laravel Version: 5.5.44
- Laravel-Datatables Version: 8.13.1

miken32:
For a `CollectionDataTable`, don't use string comparison functions when sorting two numeric values. A test case is included. Resolves #1934

### DIFF
--- a/src/CollectionDataTable.php
+++ b/src/CollectionDataTable.php
@@ -294,7 +294,15 @@ class CollectionDataTable extends DataTableAbstract
                     $first  = $a;
                     $second = $b;
                 }
-                if ($this->config->isCaseInsensitive()) {
+                if (is_numeric($first[$column] ?? null) && is_numeric($second[$column] ?? null)) {
+                    if ($first[$column] < $second[$column]) {
+                        $cmp = -1;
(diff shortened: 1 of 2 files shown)


yajra:
Released on v9.11.1. Thanks 🚀 🍻

Issue #2493, opened 2020-10-24
recordsFiltered Limit (Hard Limit in Config)


saaiful:
### Summary of problem or feature request

Laravel datatables has an interesting bug, you can request virtually unlimited rows. When using datatable in public we usually set a max record per page in order to minimize server load or to stop people from collecting all data's in a single request. 

I think adding a limit configuration in the config file can solve this issue.

PR #2494, opened 2020-10-25
Hard Limit in Config (for Pagination)

saaiful:
https://github.com/yajra/laravel-datatables/issues/2493

### DIFF
--- a/src/config/datatables.php
+++ b/src/config/datatables.php
@@ -120,9 +120,9 @@ return [
     ],
 
     /**
-     * Maximum record per page
+     * Maximum records per page
      * Set 0 for unlimited record.
      */
-    'max_record_per_page' => 100,
+    'max_records_per_page' => 100,
 
 ];
--- a/src/config/datatables.php
+++ b/src/config/datatables.php
@@ -121,7 +121,7 @@ return [
 
     /**
      * Maximum record per page
-     * Set 0 for unlimited record
+     * Set 0 for unlimited record.
      */
     'max_record_per_page' => 100,
 
--- a/src/config/datatables.php
+++ b/src/config/datatables.php
@@ -119,4 +119,10 @@ return [
         'options' => 0,
     ],
 
+    /**
+     * Maximum record per page
+     * Set 0 for unlimited record
+     */
+    'max_record_per_page' => 100,
+
 ];
--- a/src/QueryDataTable.php
+++ b/src/QueryDataTable.php
@@ -594,7 +594,9 @@ class QueryDataTable extends DataTableAbstract
      */
     public function paging()
     {
+        $max_record_per_page = $this->config->get('datatables.max_record_per_page', 0);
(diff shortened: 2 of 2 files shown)


src/config/datatables.php
```diff
@@ -119,4 +119,10 @@
         'options' => 0,
     ],
 
+    /**
+     * Maximum record per page
```
yajra, line 123: I think the correct word is records?

`Maximum records per page.`

yajra (changes requested):

yajra:
PR looks good aside from the minor typo? Will test this today on actual project. Thanks!

yajra:
Also, maybe we can also include `records_per_page` on the config while we're at it instead of hard coding it at 10?

yajra:
Just tested this and it breaks the dataTables info? Tried setting the max records to 3 and the requested length is 10.

image

Expected info should be something like  `Showing 1 to 3 of 10 results` or I might be missing something?

saaiful:
> Just tested this and it breaks the dataTables info? Tried setting the max records to 3 and the requested length is 10.
> 
> image
> 
> Expected info should be something like `Showing 1 to 3 of 10 results` or I might be missing something?

I will try to fix that.

saaiful:
I just solved the issue.
In Yajra\DataTables\Html\Builder
```/**
     * Generate DataTables js parameters.
     *
     * @param  array $attributes
     * @return string
     */
    public function parameterize($attributes = [])
    {
        $parameters = (new Parameters($attributes))->toArray();

        $max_record_per_page = $this->config->get('datatables.max_record_per_page', 0);
        if (!array_key_exists('lengthMenu', $parameters)) {
            $parameters['lengthMenu'] = [10, 25, 50, 100];
        }

        if ($max_record_per_page != 0) {
            $lengthMenu = array_unique($parameters['lengthMenu']);
            foreach ($lengthMenu as $key => $value) {
                if ($value > $max_record_per_page) {
                    unset($lengthMenu[$key]);
                }
            }
            $lengthMenu = array_values($lengthMenu);
            if (empty($lengthMenu)) {$lengthMenu = [$max_record_per_page];}
            sort($lengthMenu);
            $parameters['lengthMenu'] = $lengthMenu;
        }

        $values = [];
        $replacements = [];

        foreach (Arr::dot($parameters) as $key => $value) {
            if ($this->isCallbackFunction($value, $key)) {
                $values[] = trim($value);
                Arr::set($parameters, $key, '%' . $key . '%');
                $replacements[] = '"%' . $key . '%"';
            }
        }

        $new = [];
        foreach ($parameters as $key => $value) {
            Arr::set($new, $key, $value);
        }

        $json = json_encode($new);

        $json = str_replace($replacements, $values, $json);

        return $json;
    }
```

This will insert `lengthMenu` parameter according to `max_record_per_page` in config file. This will not affect manual `lengthMenu` insertion in "builder parameters".

PR #2500, opened 2020-10-29
made LaravelDataTables const to variable

om3rcitak:


om3rcitak:
It's directly related with this PR: https://github.com/yajra/laravel-datatables-html/pull/145

yajra:
Thanks but the html package has it own config. I think this should be included there.

Issue #2573, opened 2021-02-07
Model using polymorphic relation throw exception


warlof:
### Summary of problem or feature request

Hi, have some model using polymorphic relations that I'm calling with Datatable factory. It appears that sorting and filtering data on them are causing exceptions.

Are they not supported at time ?

This is the exception I get while doing sorting on a column :

```
[2021-02-07 21:55:22] local.ERROR: SQLSTATE[42000]: Syntax error or access violation: 1066 Not unique table/alias: 'character_wallet_transactions' (SQL: select * from `character_wallet_transactions` left join `character_wallet_transactions` on `character_wallet_transactions`.`location_id` = `character_wallet_transactions`.`` where `character_wallet_transactions`.`character_id` in (94768465) order by `character_wallet_transactions`.`name` asc limit 10 offset 0) {"userId":1,"exception":"[object] (Yajra\\DataTables\\Exceptions\\Exception(code: 0): SQLSTATE[42000]: Syntax error or access violation: 1066 Not unique table/alias: 'character_wallet_transactions' (SQL: select * from `character_wallet_transactions` left join `character_wallet_transactions` on `character_wallet_transactions`.`location_id` = `character_wallet_transactions`.`` where `character_wallet_transactions`.`character_id` in (94768465) order by `character_wallet_transactions`.`name` asc limit 10 offset 0) at /var/www/seat/vendor/yajra/laravel-datatables-oracle/src/DataTableAbstract.php:862)
[stacktrace]
#0 /var/www/seat/vendor/yajra/laravel-datatables-oracle/src/QueryDataTable.php(106): Yajra\\DataTables\\DataTableAbstract->errorResponse(Object(Illuminate\\Database\\QueryException))
#1 /var/www/seat/packages/eveseat/web/src/Http/DataTables/Common/Financial/AbstractWalletTransactionDataTable.php(94): Yajra\\DataTables\\QueryDataTable->make(true)
#2 /var/www/seat/vendor/laravel/framework/src/Illuminate/Container/BoundMethod.php(36): Seat\\Web\\Http\\DataTables\\Common\\Financial\\AbstractWalletTransactionDataTable->ajax()
#3 /var/www/seat/vendor/laravel/framework/src/Illuminate/Container/Util.php(37): Illuminate\\Container\\BoundMethod::Illuminate\\Container\\{closure}()
[...]

PR #2580, opened 2021-02-25
feat(relations): add partial polymorphic support

warlof:
make eloquent queries based on polymorphic structure searchable.

This partially address #2573


yajra:
Looks good but not sure why test fails on PHP 7.4 lowest :)

Released on v9.17.0, thanks! 🚀

warlof:
@yajra it appears I introduced a small bug for nested relationship :(

Assuming my query source model is CharacterInfo - and has a relationship called `affiliation` to the class `CharacterAffiliation`
The current change is doing `CharacterInfo::affiliation()` to retrieve relationship and determine its nature => pass

However, add a new relation to `CharacterAffiliation` called `corporation` which will target `UniverseName`.
Now, the current change will do `CharacterInfo::affiliation.corporation()` which is wrong.

I don't know how to handle that - but since the package is largely consumed, I think the best would be to rollback change until it's addressed :/

sorry about the issue :(

yajra:
No problem and thanks for the initiative to support morph relation. 🍻

Issue #2575, opened 2021-02-17
Multi_Term search disabling for some datatables


schonhoff:
### Summary of problem or feature request
Hello,

I have a problem with the multi_term search configuration. I want to provide my users the possibility in some datatables to have a multi_term search but in some datatables I want them to have a single_term search. I have no idea how to set the config "multi_term" to false just in specific datatables and not global for all datatables. Is there a possible solution for this?

This isn't working because config is a protected variable and has no getter and setter:

### Code snippet of problem

```php
    /**
     * @var EloquentDataTable $dataTable
     */
    private EloquentDataTable $dataTable;

    private function setMultiTerm(bool $multiTermSearch = true)
    {
        $this->dataTable->config->set('datatables.search.multi_term', $multiTermSearch);
    }
```

### System details

- System: Ubuntu
- PHP: 7.4
- Laravel: v8.27.0
- Laravel Datatables: v9.15.2

### Possible solution if this isn't possible atm

```php
abstract class DataTableAbstract implements DataTable, Arrayable, Jsonable
{
[...]
    /**
     * Get config
     *
     * @return \Yajra\DataTables\Utilities\Config
     */
    public function getConfig()
    {
        return $this->config;
    }
[...]
}
```

schonhoff:
Added in v9.16.0

PR #2585, opened 2021-03-17
Add set multi term on runtime

schonhoff:
This pull request adds the possibility to change the multi_term search on runtime.

See https://github.com/yajra/laravel-datatables/issues/2575

### DIFF
--- a/src/DataTableAbstract.php
+++ b/src/DataTableAbstract.php
@@ -513,7 +513,7 @@ abstract class DataTableAbstract implements DataTable, Arrayable, Jsonable
      * @param bool $multiTerm
      * @return $this
      */
-    public function setMultiTerm(bool $multiTerm = true)
+    public function setMultiTerm($multiTerm = true)
     {
         $this->config->set('datatables.search.multi_term', $multiTerm);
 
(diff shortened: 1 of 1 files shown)


schonhoff:
@yajra Should I include a test for this or is it not necessary? And can you review and add this in the upcoming version? Would be really nice!

src/DataTableAbstract.php
```diff
@@ -507,6 +507,19 @@ public function startsWithSearch($state = true)
         return $this;
     }
 
+    /**
+     * Set multi_term search config at runtime.
+     *
+     * @param bool $multiTerm
+     * @return $this
+     */
+    public function setMultiTerm(bool $multiTerm = true)
```
yajra, line 516: Maybe we should remove the `bool` type hint just to be sure this does not break on lower versions?

schonhoff: Added a commit without the type hint, so it won't break anything in lower versions. bool was included on php 7 as a type hint so maybe someone is using it on 5.x still :)

yajra:
Released on v9.16.0, thanks! 🚀

PR #2587, opened 2021-03-22
Fix check for MorphTo relation on nested relations

madpilot78:
The conditional causes errors when there are nested relations for which checking the instance of the relation method returned value causes a BadMethodCallException.

I have code that has a deep relation in the eloquent query and to test for MorphTo tries to call something like App\Models\MyModel::relation.deep.method() causing an error, since it does not exist.

I've added simple code to check the method exists before actually calling it.


madpilot78:
This patch tries to address the same issue #2588 addresses.

We have two different approaches, not sure which one is the best one.

PR #2588, opened 2021-03-22
Fix morph relation support with multi level relations

PabloCamp0s:
Recent PR https://github.com/yajra/laravel-datatables/pull/2580 fails when a multiple level relationship is used ($relation string has several dot separated relations). I've created this fix in order to check only the first relation (I don't really know how Laravel behaves with multi level relationships when all or any are morphed).


yajra:
Released on v9.17.1. Thanks! 🚀

PR #2589, opened 2021-03-22
fix: check polymorphic on nested relationship

warlof:
@yajra it appears I introduced a small bug for nested relationship :(

Assuming my query source model is CharacterInfo - and has a relationship called `affiliation` to the class `CharacterAffiliation`
The current change is doing `CharacterInfo::affiliation()` to retrieve relationship and determine its nature => pass

However, add a new relation to `CharacterAffiliation` called `corporation` which will target `UniverseName`.
Now, the current change will do `CharacterInfo::affiliation.corporation()` which is wrong.

I don't know how to handle that - but since the package is largely consumed, I think the best would be to rollback change until it's addressed :/

sorry about the issue :(

related to #2580, eveseat/seat#791

### DIFF
--- a/src/EloquentDataTable.php
+++ b/src/EloquentDataTable.php
@@ -92,7 +92,7 @@ class EloquentDataTable extends QueryDataTable
             return parent::compileQuerySearch($query, $columnName, $keyword, $boolean);
         }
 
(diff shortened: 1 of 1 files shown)


yajra:
Patch released on v9.17.1. Thanks! 🚀

PR #2606, opened 2021-04-08
Code cleanup in QueryDataTable class

schonhoff:
I discovered a warning in the orderColumn function if you use the closure syntax with Laravel .idea and PHPStorm 2021.1 installed.
Changing the PHPDoc for the function to include \Closure as a type I had no error anymore. I double checked the documentation and here the closure syntax is mentioned:
https://yajrabox.com/docs/laravel-datatables/master/order-column

Because I was in the class anyway I checked my PHPStorm warnings for the class and find 2 other things that can be cleaned up. 

Hope the PR will help

### DIFF
--- a/src/QueryDataTable.php
+++ b/src/QueryDataTable.php
@@ -309,7 +309,7 @@ class QueryDataTable extends DataTableAbstract
             }
 
             if ($this->hasFilterColumn($column)) {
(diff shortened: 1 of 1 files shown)


yajra:
Thanks, will be added on next release.

PR #2609, opened 2021-04-13
[7x] Implement setFilteredRecords feature.

kenzhegulov:
New method setFilteredRecords


src/Engines/BaseEngine.php
```diff
@@ -502,7 +502,7 @@ public function make($mDataSupport = false)
             $this->totalRecords = $this->totalCount();
 
             if ($this->totalRecords) {
-                $this->filterRecords();
+                // $this->filterRecords();
```
yajra, line 505: This should not be commented out. The logic to disable filtering should be inside the `filterRecords` method. See on latest version how it was implemented. Thanks!

Issue #2705, opened 2021-10-31
BelongsToMany relationship with orderBy returns duplicate column name created_at


dragonfly4:
### Summary of problem or feature request

Displaying list of records from Pivot Table works. Basic example with a belongsToMany in the model.
```php
$model->abilities();
```
```php
    public function abilities()
    {
        return $this->belongsToMany(Ability::class);
    }
```

Add orderBy results in duplicate column error. The Pivot Table has timestamp columns.
```php
$model->abilities()->orderBy('ability.name');
```

> error: "Exception Message:\n\nSQLSTATE[42S21]: Column already exists: 1060 Duplicate column name 'created_at' (SQL: select count(*) as aggregate from (select * from `abilities` inner join `ability_role` on `abilities`.`id` = `ability_role`.`ability_id` where `ability_role`.`role_id` = 2 order by `abilities`.`name` asc) count_row_table)"

The same command executes properly in Tinkerwell / Laravel Controller. But not via laravel-datatables. Maybe related to #1739 ?

Quick fix for me was to add a select to exclude the created_at + updated_at columns.

```php
$model->abilities()->select('id','name','label')->orderBy('ability.name');
```

### System details

- Operating System: WAMPSERVER
- PHP Version: 8.0.11
- Laravel Version: 8.68.1
- Laravel-Datatables Version: 9.18.2

yajra:
@dragonfly4 sorry for late response. I think this PR https://github.com/yajra/laravel-datatables/pull/2721 will fix the issue. Would you be able to review and give some feedbacks? Thanks!

dragonfly4:
@yajra Thanks for your reply!

I tested that but it's not fixing the issue above. Same error.

Do you want me to share anything else please?

yajra:
Let's continue the discussion in the PR. Thanks!

PR #2721, opened 2021-12-29
Don't add BelongsToMany relation column to select

affaron:
It seems like a part of "duplicate column" problems with sorting/filtering by BelongsToMany relation column are coming from this. 

When you have same column names in your main and relation tables adding a `->select()` is not helping because the relation column is still added in joinEagerLoadedColumn(). And depending on the situation either your main model column will be populated with same name relation column or you're getting an error in your aggregate queries. Example:
```
Exception Message:

SQLSTATE[42S21]: Column already exists: 1060 Duplicate column name 'name' (SQL: select count(*) as aggregate from (select `crm_companies`.`id`, `crm_companies`.`name`, `dic_company_types`.`name` from `crm_companies` left join `crm_company_types` on `crm_company_types`.`company_id` = `crm_companies`.`id` left join `dic_company_types` on `crm_company_types`.`type_id` = `dic_company_types`.`id` where LOWER(`dic_company_types`.`name`) LIKE %clearing agent% group by `crm_companies`.`name`) count_row_table)
```
(`crm_companies`.`name` is the main model column, `dic_company_types`.`name`  - automatically added relation column and hence the conflict)

So I look into the code and found this place. `$this->performJoin($table, $foreign, $other);` looks unnecessary since it's performed anyway after the switch block and `addSelect` is adding a relation column that doesn't seem to be required and causes duplicates. Please correct me if I'm wrong.

Thanks

### DIFF
--- a/src/EloquentDataTable.php
+++ b/src/EloquentDataTable.php
@@ -182,9 +182,6 @@ class EloquentDataTable extends QueryDataTable
                     $foreign = $pivot . '.' . $tablePK;
                     $other   = $related->getQualifiedKeyName();
 
-                    $lastQuery->addSelect($table . '.' . $relationColumn);
-                    $this->performJoin($table, $foreign, $other);
-
                     break;
 
                 case $model instanceof HasOneThrough:


yajra:
Per @dragonfly4 [feedback](https://github.com/yajra/laravel-datatables/issues/2705#issuecomment-1026604604), PR doesn't fixed the issue.

> I tested that but it's not fixing the issue above. Same error.

yajra:
I just recalled that sorting on belongsToMany would indeed yield to a duplicate records even when executed on the SQL level.

If you guys can provide a raw SQL on how to implement it, I will try to convert it on the package level.

src/DataTables.php
```diff
@@ -107,6 +107,7 @@ public function getConfig()
     /**
      * DataTables using Query.
      *
+
```
yajra, line 110: Please revert this change

src/DataTables.php
```diff
@@ -137,6 +138,17 @@ public function collection($collection): CollectionDataTable
         return CollectionDataTable::create($collection);
     }
 
+    /**
```
yajra, line 141: Please revert this too, already removed on v10

yajra (changes requested):

affaron:
Hey, sorry, didn't have time to get back to this. I reverted those changes from a merge I did into this branch back then. Didn't realise that it'd affect this request.

Anyway, it still seems that adding that addSelect (even the recent one in HasOneThrough case) causes some column names duplicating when you sort/search (if you already have select attached before). Maybe it's best to remove them and suggest users to use selects (it was already suggested as I recall). 

I'll try to look into this issue again later.

turdakhov:
I solved it this way: 
return $this->belongsToMany(Characteristic::class)-**>distinct()**->with('values');

Seb33300:
I think this issue is already fixed by https://github.com/yajra/laravel-datatables/pull/3238

PR #2738, opened 2022-02-07
[9.x] Fix HasOneThrough

RF1705:


yajra:
Changes looks good. Would you be able to add a test please for the use case that's being fixed here?

PR #2771, opened 2022-04-06
Changed PHPDocs to have the most specific class as their super return type

jaapieaapie1:
This allows PHPStan to understand that when you call eloquent it is now dealing with an EloquentDataTable.

Before this change it keeps assuming that it is the DataTableAbstract class and you have to fix it with a useless variable and an @var tag.

### DIFF
--- a/src/DataTables.php
+++ b/src/DataTables.php
@@ -100,7 +100,7 @@ class DataTables
      * DataTables using Query.
      *
      * @param  \Illuminate\Database\Query\Builder|mixed  $builder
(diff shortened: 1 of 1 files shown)


yajra:
Thanks! I am actually working on passing the package with phpstan via https://github.com/yajra/laravel-datatables/pull/2770 and learned something something new on your PR. If you can spare some more time, would you please check the v10 version?

yajra:
Released on v9.19.2, Thanks! 🍻

PR #2773, opened 2022-04-07
[10.x] Changed PHPDocs to have the most specific class as their super type

jaapieaapie1:
This allows PHPStan to understand that when you call eloquent it is now dealing with an EloquentDataTable.

Before this change it keeps assuming that it is the DataTableAbstract class and you have to fix it with a useless variable and an @var tag.

This is basically a copy of #2771 but for version 10.


yajra:
This would be fixed via https://github.com/yajra/laravel-datatables/pull/2770. Thanks!

PR #2787, opened 2022-05-17
Added null parameter type as allowed to handle default Action column from laravel-datatables-html

adamale:


adamale:
While upgrading laravel-datatables packages to the latest version I've encountered crashing caused by the Action column's `render` parameter being set to null initially (as defined in laravel-datatables-html). The `isJavascript()` method does not accept null values so this PR aims to fix that.

Right now I'm handling that by overriding default `render` param with the following function: `fn() => false`, but I'd definitely prefer the laravel-datatables to take care of that.

yajra:
Released on v10.0.6 🚀  Thanks a lot!

PR #2789, opened 2022-05-20
Create eloquent datatable from relation

sebdesign:
The `DataTables::eloquent()` method accepts a `Illuminate\Contracts\Database\Eloquent\Builder` which can be an eloquent builder, but also an eloquent relation.

Using `DataTables::eloquent(User::query())` works fine, but `DataTables::eloquent($user->posts())` throws the following error:

```
Exception Property [columns] does not exist on the Eloquent builder instance. 
    vendor/laravel/framework/src/Illuminate/Database/Eloquent/Builder.php:1701 Illuminate\Database\Eloquent\Builder::__get
    vendor/yajra/laravel-datatables-oracle/src/QueryDataTable.php:59 Yajra\DataTables\QueryDataTable::__construct
    vendor/yajra/laravel-datatables-oracle/src/EloquentDataTable.php:39 Yajra\DataTables\EloquentDataTable::__construct
    vendor/yajra/laravel-datatables-oracle/src/DataTableAbstract.php:167 Yajra\DataTables\DataTableAbstract::create
    vendor/yajra/laravel-datatables-oracle/src/DataTables.php:126 Yajra\DataTables\DataTables::eloquent
```

This happens because the constructor of `EloquentDataTable` passes the eloquent builder to its parent, instead of the query builder.

This PR fixes the constructor of `EloquentDataTable` so that the correct object is passed to the `QueryDataTable` constructor.

### DIFF
--- a/src/EloquentDataTable.php
+++ b/src/EloquentDataTable.php
@@ -9,6 +9,7 @@ use Illuminate\Database\Eloquent\Relations\BelongsToMany;
 use Illuminate\Database\Eloquent\Relations\HasOneOrMany;
 use Illuminate\Database\Eloquent\Relations\HasOneThrough;
 use Illuminate\Database\Eloquent\Relations\MorphTo;
+use Illuminate\Database\Eloquent\Relations\Relation;
 use Yajra\DataTables\Exceptions\Exception;
 
 /**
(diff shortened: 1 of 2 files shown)


yajra:
@sebdesign looks good and but phpstan is failing. Can you please fix? WIll also try to check later. Thanks!

sebdesign:
@yajra I've removed the default arm that throws the exception, and added a type declaration to the $model argument, instead of using phpdoc. This way PHP can enforce that the $model is a valid object, instead of ensuring this using an exception.

yajra:
Thanks! I was actually also planning to remove those exceptions when creating, glad you fixed it. Released on v10.0.7 🚀

PR #2798, opened 2022-06-21
[10.x] Make canCreate at QueryDataTable accept QueryBuilder only

mohamed-hendawy:
[canCreate](https://github.com/yajra/laravel-datatables/blob/master/src/QueryDataTable.php#L85) function at [QueryDataTable](https://github.com/yajra/laravel-datatables/blob/master/src/QueryDataTable.php) should accept `QueryBuilder` only but it accepts `EloquentBuilder`  aslo because `EloquentBuilder` contract extends `QueryBuilder `

### DIFF
--- a/src/QueryDataTable.php
+++ b/src/QueryDataTable.php
@@ -82,7 +82,7 @@ class QueryDataTable extends DataTableAbstract
      */
     public static function canCreate($source): bool
     {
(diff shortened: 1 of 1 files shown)


yajra:
Released on https://github.com/yajra/laravel-datatables/releases/tag/v10.0.8 Thanks!

PR #2816, opened 2022-07-09
[9.x] Fix HasOneThrough (added tests)

Arne1303:
I added the missing tests for the PR #2738 so that it can be merged. The original 2 commits are not my work, they belong to [RF1705](https://github.com/RF1705)

Related:
 - #2738


yajra:
Released on v9.21.2 🚀 Thanks @Arne1303  and @RF1705

Issue #2841, opened 2022-08-24
Building tables is slow on complex queries because of un-needed call of function FilteredCount() when no filter is entered. Er


marcelhoogantink:
### Summary of problem or feature request

After updating to DataTables version 10.1.2, quering of a table is slow.

Because I have some complex queries, I made a simpler (base)-query to exclude complexity not needed for the row-count.
I send this fast calculated $count to setTotalRecords().

See code below.

This worked OK in previous releases. (v9)

Now the queries are slow again. 
Looking into it, shows that this is slowed down, by calling function filteredCount(), but I don't have a filter.

### Code snippet of problem

```php
    public function dataTable($query)
    {

        // simpler count() for speeding-up build of the view (orginal count() is slow because of taking the full complexity of the query with multiple joins
        $count= self::defQueryBase()->count();

        $dataTable = datatables($query)
            ->setTotalRecords($count)
            ->addColumn(Def::VIEW__EAN_RECORDS__ACTION,
                function ($row)
                {
                    return $this->fillActionColumn($row);
                });


        return $dataTable;
    }
```
### System details

- Operating System: Windows 10
- PHP Version: 8.0.18
- Laravel Version: 9.25.1
- Laravel-Datatables Version 10.1.2

yajra:
Thanks for reporting, the filter count was added back at https://github.com/yajra/laravel-datatables/pull/2778 to fix several issues. Will try to review the implementation again, but if you can, please do not hesitate to submit a PR.

marcelhoogantink:
Tried to make a PR

marcelhoogantink:
Closing this issues due to no interest by the owner, showed by ignoring comments in the PR.

PR #2852, opened 2022-09-06
New check for applied filter (#2841)

marcelhoogantink:
Here is my first PR for issue #2841

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,9 @@
 /coverage
 composer.phar
 composer.lock
+.idea/.gitignore
+.idea/datatableClone.iml
+.idea/deployment.xml
+.idea/modules.xml
+.idea/php.xml
+.idea/vcs.xml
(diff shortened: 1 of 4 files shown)


yajra:
@marcelhoogantink great PR, almost there 🎊 . 

Can you please fix the failing static analysis and tests? StyleCI can be fixed later on and can be ignored. Thanks!

marcelhoogantink:
The error in the static analysis are triggered on code I did not touch.
The last 3 errors in the tests are errors on tests that have a higher filteredRecords than totalRecords, so this test seems not logical to me.

yajra:
Please fix at least the PHPUnit test for this to be merged. Thanks!

marcelhoogantink:
Ok, I tried, but in my aproach, when there is no filter active, the recordsFiltered is equal to the recordsTotal.
In this situation the tests 'it_can_set_total_records' and 'it_can_set_zero_total_records' fails because the recordsFiltered is not relevant here.
Please reply ...

marcelhoogantink:
To bad, no comment after may last question.
So no ibterersts?
I am closing the PR.

PR #2861, opened 2022-10-03
Methods pointing to the "uncustomizable" classes.

OzanKurt:
This allows people to use corresponding method calls like:

```
$builder = datatables()->eloquent($query);
```

instead of needing to use the `make` method.

```
$builder = datatables()->make($query);
```

Since only the `make` is using the classes from `datatables.engines`, the code is not consistent.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -23,7 +23,8 @@
   },
   "require-dev": {
     "nunomaduro/larastan": "^2.1",
-    "orchestra/testbench": "^7.3"
+    "orchestra/testbench": "^7.3",
+    "yajra/laravel-datatables-html": "^9.0"
   },
   "suggest": {
     "yajra/laravel-datatables-buttons": "Plugin for server-side exporting of dataTables.",
--- a/composer.json
+++ b/composer.json
@@ -23,8 +23,7 @@
   },
   "require-dev": {
     "nunomaduro/larastan": "^2.1",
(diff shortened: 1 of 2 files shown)


OzanKurt:
Done

yajra (commented):
phpstan is failing since the class is now being fetched from config. There is a chance that the package will fail if the config is wrong.

```
<file name="src/DataTables.php">
  <error line="117" column="1" severity="error" message="Cannot call static method create() on mixed." />
  <error line="130" column="1" severity="error" message="Cannot call static method create() on mixed." />
  <error line="[14](https://github.com/yajra/laravel-datatables/actions/runs/3254410336/jobs/5342627471#step:5:15)3" column="1" severity="error" message="Cannot call static method create() on mixed." />
</file>
```

Maybe we can ignore this for it to pass? https://phpstan.org/user-guide/ignoring-errors#ignoring-in-code-using-phpdocs

```
/** @phpstan-ignore-next-line */
```

OzanKurt:
I'm sorry mate, I can't deal with these nonsense phpstan errors. I even tried installing the package standalone and it's just too much waste of time.

image

yajra:
Looks good and thanks for the hassle. Will review this as soon as I can.

yajra:
Released on v10.2.0, thanks a lot! 🚀

PR #2901, opened 2022-11-15
found a scenario where the data table search box can't search the val…

Sujon94:
…ue consisting of '&' though the data exists in the table. Fixed the issue.

### DIFF
--- a/src/DataTableAbstract.php
+++ b/src/DataTableAbstract.php
@@ -737,8 +737,8 @@ abstract class DataTableAbstract implements DataTable
     public function filtering(): void
     {
         $keyword = $this->request->keyword();
-        /**Following line of code to convert & or any other special character to html string. **/
-        $keyword = (is_string($keyword) || $keyword instanceof Htmlable) ? e($keyword) : $keyword;
+        /**Following line of code to convert `&` or any other special character to html string. **/
+        $keyword = e($keyword);
 
         if ($this->config->isMultiTerm()) {
             $this->smartGlobalSearch($keyword);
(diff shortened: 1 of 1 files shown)


src/DataTableAbstract.php
```diff
@@ -737,6 +737,8 @@ protected function filterRecords(): void
     public function filtering(): void
     {
         $keyword = $this->request->keyword();
+        /**Following line of code to convert & or any other special character to html string. **/
+        $keyword = (is_string($keyword) || $keyword instanceof Htmlable) ? e($keyword) : $keyword;
```
yajra, line 741: `$keyword` is always a string thus Htmlable is not needed. Maybe we can just convert to HTML string directly from the `request->keyword()` method?

yajra (changes requested):

src/DataTableAbstract.php
Sujon94: Yes, you are right. We cam ommit the conditional part and directly convert to HTML string.

Sujon94: I have commited. Please take a review.

yajra:
The changes look good. However, this only changes the filtering keyword. Doing a column search would yield the previous results. 

Would you be able to review this PR https://github.com/yajra/laravel-datatables/pull/2903? That PR will make the keyword to HTML entities anywhere it is used.

Sujon94:
Ok, I will review.

yajra:
Closing in favor of https://github.com/yajra/laravel-datatables/pull/2903.

PR #2903, opened 2022-11-17
fix: encode search keyword

yajra:
Fix #2901


yajra:
Note: need to review as encoding the keyword is a change in behavior and might cause a breaking change.

src/Utilities/Request.php
```diff
@@ -194,10 +194,10 @@ public function columnKeyword(int $index): string
     protected function prepareKeyword(float|array|int|string $keyword): string
     {
         if (is_array($keyword)) {
-            return implode(' ', $keyword);
+            $keyword = implode(' ', $keyword);
         }
 
-        return (string) $keyword;
+        return e((string) $keyword);
```
Arkhas, line 200: casting into a string is no longer needed as `e()`function return a string
```suggestion
        return e($keyword);
```

yajra: I just recalled that the cast was to pass phpstan since the keyword param also accepts int & float values. https://github.com/yajra/laravel-datatables/actions/runs/3519589634/jobs/5899694603

dyaskur (changes requested):
This change will make we can not search non encoded special chars. an example we have data "Hand & Soulder"(no encode) and we search using "&", and the data will not shown.

My suggestion is using attribute like searchable in datatable options.

an example:

                  {data: 'name', name: 'name',searchable: true, is_encoded:true},

and if `is_encoded:true`, we use `e()` function, otherwise return without encode.

yajra:
@dyaskur thanks for the review. However, adding `is_encoded:true` in the script will not affect the serverSide implementation as the js library will not send it back in the ajax request.

How about if we add another control on a column level that toggles encoded search? Maybe something like:

```php
->encodedSearch(['col1', 'col2'])
```

dyaskur:
I see, I thought datatable will send all column attribute to the ajax request. If can't send by client side, we need to set it on server side. And I think your idea looks a good idea.

aravael:
@yajra , recently came up with a similar solution while debugging a weird behavior with the `collection` as a source. This change adds equality to a search input and modified collection at `DataProcessor escapeRow` which uses laravel's `e()` func. 
Any estimates on when this will be merged?

yajra:
@aravael, would you be able to test this and see the impact on your existing project? Would you approve this PR?

aravael:
Hi @yajra . Recently managed to test it, can't approve this patch. Though it fixes the collection search, it breaks the `builder source`. Since we sanitize the input with `e()`, the database data are still not modified, therefore not found the result.

It works with a collection because both the `collection ` and `input` sanitized. In case of builder source it doesn't work that way. Need to process that out.

PR #2906, opened 2022-11-21
fix: Custom Order on eager loaded relationships was not working

Arkhas:
Following this issue : https://github.com/yajra/laravel-datatables/discussions/2905

When creating custom order for a column with dot notation, 

if the column name was `relation.column` (`user.id`)
AND the relationship was eagerloaded, 
the default order would be `relation_table.column` (`users.id`)

BUT if we would customise the order of that column like that : 

```
->orderColumn('user.id', function ($query, $order) {
    // For testing purposes, revert the order
    $query->orderBy('users.id', $order == 'desc' ? 'asc' : 'desc');
})
```

the custom order would not be applied

by checking if a custom order exists for our column definition name, we fix that behavior

### DIFF
--- a/src/QueryDataTable.php
+++ b/src/QueryDataTable.php
@@ -633,7 +633,7 @@ class QueryDataTable extends DataTableAbstract
 
                 if ($this->hasOrderColumn($orderable['name'])) {
                     $this->applyOrderColumn($orderable['name'], $orderable);
-                } else if ($this->hasOrderColumn($column)) {
+                } elseif ($this->hasOrderColumn($column)) {
                     $this->applyOrderColumn($column, $orderable);
                 } else {
                     $nullsLastSql = $this->getNullsLastSql($column, $orderable['direction']);
--- a/src/QueryDataTable.php
+++ b/src/QueryDataTable.php
@@ -631,7 +631,9 @@ class QueryDataTable extends DataTableAbstract
             ->each(function ($orderable) {
                 $column = $this->resolveRelationColumn($orderable['name']);
 
-                if ($this->hasOrderColumn($column)) {
+                if ($this->hasOrderColumn($orderable['name'])) {
+                    $this->applyOrderColumn($orderable['name'], $orderable);
+                } else if ($this->hasOrderColumn($column)) {
                     $this->applyOrderColumn($column, $orderable);
                 } else {
                     $nullsLastSql = $this->getNullsLastSql($column, $orderable['direction']);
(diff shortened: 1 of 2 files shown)


Arkhas:
@yajra I don't know how to reduce code duplication without reducing consistency with the other tests, where would you put them ?

yajra:
Code duplication is because of tests that should be acceptable. I am still figuring out how to configure it on SonarCloud. That check can be ignored.

Anyways, PR looks good. Will review this as soon as I can. Thanks!

src/QueryDataTable.php
```diff
@@ -631,7 +631,9 @@ protected function defaultOrdering(): void
             ->each(function ($orderable) {
                 $column = $this->resolveRelationColumn($orderable['name']);
 
-                if ($this->hasOrderColumn($column)) {
+                if ($this->hasOrderColumn($orderable['name'])) {
```
yajra, line 634: Having some doubts about what will happen if the relationship is not joined in the query. Maybe we should switch the check so it will be a fallback when the eager loading check failed?

```PHP
                if ($this->hasOrderColumn($column)) {
                    $this->applyOrderColumn($column, $orderable);
                } elseif ($this->hasOrderColumn($orderable['name'])) {
                    $this->applyOrderColumn($orderable['name'], $orderable);
```

Arkhas: if the relation is not eager load, for an EloquentDatatable, it would return the column name directly : 

EloquentDataTable:157

```
if ($this->isNotEagerLoaded($relation)) {
        return $column;
}
```

so  as `$column == $orderable['name']` there is no difference between the if

And the other hand even if the relation **IS**  join the query,I think that a custom order should always be applied (if not, why specified a custom orderColumn()) because if you don't want to join a table for any reason for your order, you would not be able to do that	

exemple : 

```php
protected function setUp(): void
    {
        parent::setUp();

        $this->app['router']->get('/relations/belongsTo', function (DataTables $datatables) {
            return $datatables->eloquent(Post::select('posts.*')
											// without eagerload
                                             //->with('user')
                                             )
                              ->orderColumn('user.id', function ($query, $order) {
								// Order with a subquery instead of a join
                                  return $query->orderBy(DB::raw('(select id from users where users.id = posts.user_id)'),  $order == 'desc' ? 'asc' : 'desc');
                              })
                              ->editColumn('user.id', function ($post) {
                                  return $post->user->id;
                              })
                              ->toJson();
        });
    }
```

Arkhas:
bump on this

yajra:
Sorry, currently away from Laravel for quite some time. Changes look good and tests are passing. I think this is safe to release. Thanks!

yajra:
Released on v10.2.3 🚀 Thanks!

PR #2981, opened 2023-03-30
feat: Expose autoFilter setter to disable post filtering for already filter…

thureos:
…ed datasets

I had a problem when using the collection one so the already filtered dataset coming from an api was performing extra filtering and thus not showing the right results. This fixed this problem

### DIFF
--- a/src/DataTableAbstract.php
+++ b/src/DataTableAbstract.php
@@ -583,7 +583,7 @@ abstract class DataTableAbstract implements DataTable
 
         return $this;
     }
-    
+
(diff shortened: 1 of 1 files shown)


yajra (commented):
This looks good to go, however, would you be able to add a test for this, please?

Thanks!

yajra:
Sorry for overlooking this PR. Good to go, thanks!

yajra:
Released on v10.6.0 🚀 thanks!

PR #3003, opened 2023-05-29
Fix return type for `setTransformer()` and `setSerializer`

sebdesign:
This PR https://github.com/yajra/laravel-datatables/pull/1383 added the `DataTableAbstract` return type for `setTransformer()` and `setSerializer` methods. This doesn't totally fix the IDE autocompletion, because the IDE cannot autocomplete methods from the actual datatable class that the method was called, it can't autocomplete methods from the `DataTableAbstract` class only. Returning `static` tells the IDE to autocomplete methods from the class that the method was called.


yajra:
Released on v10.4.2 🚀 Thanks!

PR #3007, opened 2023-06-05
Prevent the filteredCount() query if no filter is applied to the initial query

### LINKED ISSUE #2959: Datatables Eloquent Triggering count(*) query twice
Updated Datatables to latest version recently but noticed that the behaviour has changed in such a way that count(*) queries are triggered twice?

Example Code:
$response= datatables()->of($query)->toJson();

SQL Queries:
select count(*) as aggregate from (XXX);
select count(*) as aggregate from (XXX);
[Actual SELECT query]

The issue here is that we are dealing for fairly large datasets (in millions) and count(*) statements are quite expensive. Moreover the "XXX" statement is the same for both count(*) queries. 

Is there anyway to prevent this double query?

Arkhas:
To reduce the processing made by the count on the filtered query, I performed a check if the query has been altered after the initial query

It will improve performance on the first call to a DataTable (as no filter has been applied yet, and any subsequent request without a search)

I had to modify some assertions in the tests as the filteredRecords should never be greater than the totalRecords.

fixes #2959

### DIFF
--- a/src/QueryDataTable.php
+++ b/src/QueryDataTable.php
@@ -49,13 +49,13 @@ class QueryDataTable extends DataTableAbstract
     protected bool $keepSelectBindings = false;
 
     /**
-     * @param QueryBuilder $builder
+     * @param  QueryBuilder  $builder
      */
     public function __construct(QueryBuilder $builder)
     {
-        $this->query   = $builder;
+        $this->query = $builder;
         $this->request = app('datatables.request');
-        $this->config  = app('datatables.config');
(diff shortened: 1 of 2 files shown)


yajra:
Looks good, will check this as soon as I can. Thanks!

hpacleb:
Looks good to me, my initial load query decreased by 1 and filtered datatables still works as is.

yajra:
Released on https://github.com/yajra/laravel-datatables/releases/tag/v10.4.3 🚀 Thanks!

PR #3008, opened 2023-06-05
feat: Optimize countQuery with complex select

Arkhas:
When performing subqueries, some of the keywords used to determine if a query is complexe can be found in subqueries inside the main query : 

```
protected function isComplexQuery($query): bool
{
     return Str::contains(Str::lower($query->toSql()), ['union', 'having', 'distinct', 'order by', 'group by']);
}
```

For exemple : 

```sql
select 
    "users".*,
    (select id from posts where posts.user_id = users.id order by created_at desc limit 1) as last_post_id
from "users"
```

This query was detected as complexe, but the subquery had no impact on the number of records returns, 
The subquery was executed and had a huge impact on performance

To force this behavior, you can use the `ignoreSelectsInCountQuery()` on a QueryDatatable : 

```php
 $dataTable = app('datatables')->of(
            DB::table('users')
                ->select('users.*')
                ->addSelect([
                    'last_post_id' => DB::table('posts')
                                        ->whereColumn('posts.user_id', 'users.id')
                                        ->orderBy('created_at')
                                        ->select('id'),
                ])
                ->orderBy(DB::table('posts')->whereColumn('posts.user_id', 'users.id')->orderBy('created_at')->select('created_at')
                )
        )->ignoreSelectsInCountQuery();
```

### DIFF
--- a/src/QueryDataTable.php
+++ b/src/QueryDataTable.php
@@ -158,7 +158,7 @@ class QueryDataTable extends DataTableAbstract
 
         if ($this->isComplexQuery($builder)) {
             $builder->select(DB::raw('1'));
-            
+
             if (! $this->isComplexQuery($builder)) {
                 return $this->getConnection()
                     ->query()
@@ -167,6 +167,7 @@ class QueryDataTable extends DataTableAbstract
             }
 
             $builder = clone $this->query;
+
             return $this->getConnection()
                 ->query()
                 ->fromRaw('('.$builder->toSql().') count_row_table')
--- a/src/QueryDataTable.php
+++ b/src/QueryDataTable.php
@@ -810,10 +810,10 @@ class QueryDataTable extends DataTableAbstract
     }
 
    /**
-     * Ignore the selects in count query.
-     *
-     * @return $this
-     */
+    * Ignore the selects in count query.
+    *
+    * @return $this
+    */
    public function ignoreSelectsInCountQuery()
    {
        $this->ignoreSelectInCountQuery = true;
--- a/src/QueryDataTable.php
+++ b/src/QueryDataTable.php
@@ -164,16 +164,15 @@ class QueryDataTable extends DataTableAbstract
         $builder = clone $this->query;
 
         if ($this->isComplexQuery($builder)) {
-            $builder->select(DB::raw('1'));
+            $query_without_selects = clone $this->query;
+            $query_without_selects->select(DB::raw('1'));
 
-            if ($this->ignoreSelectInCountQuery || ! $this->isComplexQuery($builder)) {
+            if ($this->ignoreSelectInCountQuery || ! $this->isComplexQuery($query_without_selects)) {
                 return $this->getConnection()
                     ->query()
-                    ->fromRaw('('.$builder->toSql().') count_row_table')
-                    ->setBindings($builder->getBindings());
+                    ->fromRaw('('.$query_without_selects->toSql().') count_row_table')
+                    ->setBindings($query_without_selects->getBindings());
             }
-
-            $builder = clone $this->query;
 
             return $this->getConnection()
                 ->query()
--- a/src/QueryDataTable.php
+++ b/src/QueryDataTable.php
@@ -57,13 +57,13 @@ class QueryDataTable extends DataTableAbstract
     protected bool $ignoreSelectInCountQuery = false;
 
     /**
-     * @param QueryBuilder $builder
+     * @param  QueryBuilder  $builder
      */
     public function __construct(QueryBuilder $builder)
     {
-        $this->query   = $builder;
+        $this->query = $builder;
         $this->request = app('datatables.request');
-        $this->config  = app('datatables.config');
+        $this->config = app('datatables.config');
         $this->columns = $builder->columns;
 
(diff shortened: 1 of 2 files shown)


src/QueryDataTable.php
```diff
@@ -156,10 +157,21 @@ public function prepareCountQuery(): QueryBuilder
         $builder = clone $this->query;
 
         if ($this->isComplexQuery($builder)) {
+            $builder->select(DB::raw('1'));
+
+            if (! $this->isComplexQuery($builder)) {
```
yajra, line 162: Can you please review and fix this phpstan findings? Thanks!

yajra (changes requested):

src/QueryDataTable.php
Arkhas: well I'm not sure how to fix this as PHPStan is wrong

(you can test this by commenting the inside of the if and running the `test_simple_queries_with_complexe_select_are_wrapped_without_selects` test, it will fail)

Arkhas:
ping @yajra

src/QueryDataTable.php
```diff
@@ -48,6 +49,13 @@ class QueryDataTable extends DataTableAbstract
      */
     protected bool $keepSelectBindings = false;
 
+    /**
+     * Flag to ignore the selects in count query.
+     *
+     * @var bool
+     */
+    public bool $ignoreSelectInCountQuery = false;
```
yajra, line 57: Can we make this protected?

```suggestion
    protected bool $ignoreSelectInCountQuery = false;
```

src/QueryDataTable.php
```diff
@@ -156,10 +164,21 @@ public function prepareCountQuery(): QueryBuilder
         $builder = clone $this->query;
 
         if ($this->isComplexQuery($builder)) {
+            $builder->select(DB::raw('1'));
+
+            if ($this->ignoreSelectInCountQuery || ! $this->isComplexQuery($builder)) {
```
yajra, line 169: Maybe we remove and fix the negated boolean?

```suggestion
            if ($this->ignoreSelectInCountQuery) {
```

yajra (changes requested):

src/QueryDataTable.php
yajra: Oh, test failed when it was removed... checking further.

tests/Unit/QueryDataTableTest.php
```diff
@@ -28,9 +28,69 @@ public function test_complex_query_are_wrapped_and_countable()
...
+              ->addSelect([
+                  'last_post_id' => DB::table('posts')
+                                      ->whereColumn('posts.user_id', 'users.id')
+                                      ->orderBy('created_at')
+                                      ->select('id'),
+              ])
+        );
+
+        $this->assertQueryWrapped(true, $dataTable->prepareCountQuery());
+        $this->assertQueryHasNoSelect(true, $dataTable->prepareCountQuery());
```
yajra, line 90:
```suggestion
        $this->assertQueryHasNoSelect(false, $dataTable->prepareCountQuery());
```

Should use `ignoreSelectsInCountQuery` if want to have no select as true?

src/QueryDataTable.php
yajra: Test fixed by setting false in has no select.

tests/Unit/QueryDataTableTest.php
Arkhas: If the query does not need to use the selects, it remove them, the `ignoreSelectsInCountQuery()` is use only to force this behavior (for example, when performing a subquery inside an order on the main query as in the test)

src/QueryDataTable.php
Arkhas: # It seems that PHPStan does not understand that this change the `$builder` instance : 
```php
$builder->select(DB::raw('1'));
```

We are first checking if the query with the select is a complexe query,

if true then

we are overriding the selects, to only select '1', removing everything.

then, rechecking if the query is still complexe (if anything outside the selects use a order, union, group by etc...)

If the query is no more complexe, we use the new query without selects, else, we use it with the selects as the order, group by etc... might use them

Arkhas:
I am struggling every time with the CI, does StyleCI have an automated fixer ? It would be much easier :)

yajra:
> I am struggling every time with the CI, does StyleCI have an automated fixer ? It would be much easier :)

No problem with StyleCI, I can automate the PR afterwards.

yajra:
There is still a Static Analysis issue:

```php
 ------ --------------------------------------------- 
  Line   QueryDataTable.php                           
 ------ --------------------------------------------- 
  168    Negated boolean expression is always false.  
 ------ --------------------------------------------- 
```

But let me check again further. If it's a false positive, maybe we can just add it to ignored errors?

Arkhas:
yeah I don't understand why it is detected like that

Arkhas:
where you able to check it ? Or do you need more explanation on what the function does ?

yajra (approved):

yajra:
Released on v10.4.4 🚀 including the fixes for phpstan. Thanks!

PR #3009, opened 2023-06-06
feat: add ability to disable eloquent getter mutator

Arkhas:
The issue was when a model have a getter overwriting an existing attribute in the database (See the new User Migration and User Model)

we have now a "color" column and a getColorAttribute() which override the value if null.

For simplicity, in the test, there is no real issue with it, but if you have this type of getter (poor code quality, but it is for the sake of the example): 

```php
public function getColorAttribute(){
	return $this->posts->count() ? '#000000' : '#ff0000'
}
```

with this query : 

```php
User::select('users.*')
```

This would result in a n+1 as every User row will call this getter and make a query to the posts table for each row

Facing this issue, you have multiple solution : 

Eager loading your posts (even if not needing it): 

```php
User::with('posts')->select('users.*')

//	If you want to create a Table of post your query would have to be like : 

Post::with('user.posts')->select('posts.*')
```

Not selecting the problematics columns : 

```php
User::select('users.id', 'users.name', 'users.email')
```

So, I propose to add a key in the config `ignore_getters` false, by default to keep the actual behavior

### DIFF
--- a/tests/Integration/IgnoreGettersTest.php
+++ b/tests/Integration/IgnoreGettersTest.php
@@ -4,7 +4,6 @@ namespace Yajra\DataTables\Tests\Integration;
 
 use Illuminate\Foundation\Testing\DatabaseTransactions;
 use Yajra\DataTables\DataTables;
-use Yajra\DataTables\Tests\Models\Post;
 use Yajra\DataTables\Tests\Models\User;
 use Yajra\DataTables\Tests\TestCase;
 
@@ -41,7 +40,7 @@ class IgnoreGettersTest extends TestCase
         $this->assertCount(20, $response->json()['data']);
     }
 
-     /** @test */
+    /** @test */
     public function it_ignore_the_getter_value_with_ignore_getters_config()
     {
         app('config')->set('datatables.ignore_getters', true);
--- a/tests/Integration/IgnoreGettersTest.php
+++ b/tests/Integration/IgnoreGettersTest.php
@@ -29,7 +29,7 @@ class IgnoreGettersTest extends TestCase
     }
 
     /** @test */
-    function it_return_the_getter_value_without_ignore_getters_config()
+    public function it_return_the_getter_value_without_ignore_getters_config()
     {
         $response = $this->call('GET', '/ignore-getters');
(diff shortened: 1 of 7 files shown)


yajra:
Not yet, as the last time I checked, this was in draft. Is it ready for review? Will try to check asap.

Arkhas:
@yajra were you able to check on this one ? :)

Arkhas:
ping @yajra :)

yajra:
Can you please fix the conflicts? Thanks and apologies for the late response. A bit busy at work :)

Arkhas:
Sure no problem ;)

tests/Models/User.php
```diff
@@ -27,4 +27,9 @@ public function user()
     {
         return $this->morphTo();
     }
+
+    public function getColorAttribute()
+    {
+        return $this->color ?? '#000000';
```
yajra, line 33: I think it would be nice if we can have an assertion on the attribute getter to ensure that `#000000` is returned when the color is null.

yajra (commented):
Updates look good and it would be nice if we have more assertions on the getter attributes but it can be done some other time.

Will check further this week and test on an actual project. But this looks good to go. Thanks!

tests/Models/User.php
Arkhas: done 👍

tests/Integration/IgnoreGettersTest.php
```diff
@@ -0,0 +1,70 @@
...
+                'email' => 'foo@bar.com',
+                'color' => null
+            ]);
+
+        $this->assertEquals('#000000', $user->color);
+        $this->assertEquals('#000000', $user->refresh()->toArray()['color']);
+    }
+
+    /** @test */
+    public function it_return_the_getter_value_without_ignore_getters()
```
yajra, line 28: Sorry for not clearing my request. The assertion is on this test that when ignore_getters are not set, the computed attribute would still be displayed.

yajra:
Released in https://github.com/yajra/laravel-datatables/releases/tag/v10.7.0 🚀 Thanks!

PR #3018, opened 2023-06-20
feat: Prevent editColumn when column is not shown

Arkhas:
To match the behavior of the filterColumn in globalSearch where only the displayed columns are searched, I propose the following : 

Example : a POST Datatable
When doing this 

```php
->editColumn('user.name', function ($post) {
	return "<a href='/users/{$post->user->id}'>{$post->user->name}</a>";
})
```

If you want to reuse this Datatable on the User page, you would probably do this : 

```php
$columns = [
	'user.name' => ['title' => 'Author'],
	'title' => ['title' => 'Post Title'],
	'comments_count' => ['title' => '#Comments'],
];

if($this->user){
	array_forget($columns, 'user.name');
}
```

and in your query, you don't need to eagerload the user when already in contexte : 

```
Post::query()
->withCount('comments')
->when(!$this->user, function($query){
	$query->with('user:id,name');
});
```

that way, you would be able to reuse this DataTable either Globaly or In the context of a User  

BUT 

with the editColumn, it will lazy load the `user` relationship for every post, so you would have to do something like that too : 

```php
if(!$this->user){
	$datatable->editColumn('user.name', function ($post) {
		return "<a href='/users/{$post->user->id}'>{$post->user->name}</a>";
	});
}
```

There is other use cases when you are not in any contexte, but only depend on the payload of your DataTable like manually calling the controller with a get Payload like in the Feature Test : 

```php
$this->call('GET', '/query/edit-columns', [
            'columns' => [
                ['data' => 'name', 'name' => 'name', 'searchable' => 'true', 'orderable' => 'true'],
            ],
        ])
```

So instead, now it would only execute the editColumn IF the column is in the payload OR if now columns are passed in the payload

I had to make the following check : `!count($this->request->columns())` to ensure that calling without any columns like in the tests would not breaks all the tests, but I would recommend fixing the tests instead as there should be no request without a columns payload

### DIFF
--- a/src/DataTableAbstract.php
+++ b/src/DataTableAbstract.php
@@ -231,7 +231,9 @@ abstract class DataTableAbstract implements DataTable
      */
     public function editColumn($name, $content): static
     {
-        $this->columnDef['edit'][] = ['name' => $name, 'content' => $content];
+        if (!count($this->request->columns()) || in_array($name, Arr::pluck($this->request->columns(), 'name'))) {
+            $this->columnDef['edit'][] = ['name' => $name, 'content' => $content];
+        }
 
         return $this;
     }
--- a/src/DataTableAbstract.php
+++ b/src/DataTableAbstract.php
@@ -152,7 +152,7 @@ abstract class DataTableAbstract implements DataTable
     /**
      * Can the DataTable engine be created with these parameters.
      *
-     * @param mixed $source
+     * @param  mixed  $source
      * @return bool
      */
     public static function canCreate($source)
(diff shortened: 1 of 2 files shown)


Arkhas:
@yajra were you able to check this ? :)

yajra:
The changes look good. However, I think this is a change in behavior that might cause a breaking change for others.

Example: 

We sometimes edit a column but do not display it in the table. It is displayed via custom javascript or after some user interaction.

Arkhas:
I made the following change : 

You can now use the `editOnlySelectedColumns ()` on your DataTable.

Any `editColumn()` made before would still be edited
Any  `editColumn()` made after would not be edited

yajra (approved):
Approved. Maybe you can also send a PR in the [docs](https://github.com/yajra/laravel-datatables-docs) as it's a bit confusing to use. Thanks!

yajra:
Release on v10.5.0 🚀 Thanks!

Arkhas:
https://github.com/yajra/laravel-datatables-docs/pull/93 here :)

Issue #3010, opened 2023-06-08
Error Thrown by DataTableAbstract::getColumnNameByIndex()  on Laravel 10


fahmidyt:
### Summary of problem or feature request

When I try to update Laravel 9 -> 10. There is a bug shown. And it is caused when the method `->toJson` is called. The reason behind that is that I use the `DB::raw()` on my Query Builder and it caused the `DataTableAbstract::getColumnNameByIndex()` thrown error. The error was basically caused by type error. Details on the following snippets.

### Code snippet of problem

The following snippets are from my controller.
```php
class Controller extends MYController {
    public function read(Request $request) {
        $queryBuilder = DB::table('tbl_association')->select('id', 'association_name');
        // this line will cause the bug
        $queryBuilder->addSelect(DB::raw('(SELECT 1)'));

        return DataTables::query($queryBuilder)->toJson();
    }
}
```

And the following snippets are the error has thrown
```json
{
    "message": "Yajra\\DataTables\\DataTableAbstract::getColumnNameByIndex(): Return value must be of type string, Illuminate\\Database\\Query\\Expression returned",
    "exception": "TypeError",
    "file": "C:\\folder\\to\\project\\vendor\\yajra\\laravel-datatables-oracle\\src\\DataTableAbstract.php",
    "line": 1030,
    "trace": ["..."]
}
```

And I also read this article [https://laravel.com/docs/10.x/upgrade#database-expressions](https://laravel.com/docs/10.x/upgrade#database-expressions)
There's a thing that might be related to this

```
Notably, the grammar's raw string value must now be retrieved via the expression's getValue(Grammar $grammar) method. Casting an expression to a string using (string) is no longer supported.
```

And I hope you can help with this.. Thanks!

### System details

- Operating System: Windows 11
- PHP Version: 8.1.10
- Laravel Version: 10.13.2
- Laravel-Datatables Version: 10.1.0

Nacoma:
Still an ongoing issue.

yajra:
Fixed vi #3029, v10.6.2. Thanks!

PR #3029, opened 2023-07-13
#3010 - convert expressions to strings

Nacoma:
#3010 

The query `Grammar` is required to stringify the expression. This ends up needing to be changed in both files as the `AbstractDataTable` does not have access to a connection to retrieve the grammar.

### DIFF
--- a/src/EloquentDataTable.php
+++ b/src/EloquentDataTable.php
@@ -10,6 +10,7 @@ use Illuminate\Database\Eloquent\Relations\HasOneOrMany;
 use Illuminate\Database\Eloquent\Relations\HasOneThrough;
 use Illuminate\Database\Eloquent\Relations\MorphTo;
 use Illuminate\Database\Eloquent\Relations\Relation;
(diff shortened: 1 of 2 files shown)


src/EloquentDataTable.php
```diff
@@ -140,6 +141,19 @@ protected function isMorphRelation($relation)
         return $isMorph;
     }
 
+    protected function getColumnNameByIndex(int $index): string
```
yajra, line 144: I believed this method can be removed since it will be inherited from the QueryDataTable.

yajra (changes requested):

src/EloquentDataTable.php
Nacoma: This is called indirectly from the `CollectionDataTable` as well. I opted to not make it abstract and move the implementation because it will break any userland implementations.

yajra (approved):

yajra:
Released on v10.6.2 🚀 Thanks!

PR #3035, opened 2023-07-24
feat: Ability to use deep relations for searching

OzanKurt:
So the idea is that we have `Post -> belongsTo -> User -> belongsToMany -> Role` relationships.

Currently when we enter the following column data:

```
[
    'data' => 'user.roles.role', 
    'name' => 'user.roles.role', 
    'searchable' => 'true',
    'orderable' => 'true',
],
```

We get an error `roles.role` column not found, since it only checks for 1 relation, which is the `user` relation in out case.

From now on if there are more than 1 `.` character it will check the other relation as a nested `whereHas` clouse

### DIFF
--- a/src/EloquentDataTable.php
+++ b/src/EloquentDataTable.php
@@ -81,13 +81,35 @@ class EloquentDataTable extends QueryDataTable
     /**
      * @inheritDoc
      */
-    protected function compileQuerySearch($query, string $column, string $keyword, string $boolean = 'or'): void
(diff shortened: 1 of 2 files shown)


OzanKurt:
@yajra Could you merge this if possible?

yajra (approved):

yajra:
Released in https://github.com/yajra/laravel-datatables/releases/tag/v10.7.0 🚀 Thanks!

PR #3082, opened 2023-10-11
feat: Scout Search Implementation

### LINKED ISSUE #3077: Scout / Meilisearch Support
I have implemented a custom solution for scout/meilisearch support in my datatables around a month ago. Here are some explanations of it so you get the idea. If there is interest in this, i can clean this up, try to make it generally compatible to all scout engines (currently only meilisearch) and create a pull request @yajra 

The core idea was to keep the normal search/ordering/filter abilities and add the meilisearch engine on top of it so both can be used together. If the meilisearch engine is not indexed yet or has a downtime, the system automatically falls back to the normal search engine. Additionally it is possible to define both meilisearch filters and normal filters so that you can join and filter other tables/cols after the meilisearch search was already done.

In my case there are static products (that are indexed to meilisearch and searched via meilisearch), but these products had dynamic prices that were changing too often to index them properly. So the search engine first searches for all products matching the keyword and then adds a mysql filter on top of them.

Example usage:
```php
public function dataTable(QueryBuilder $query): EloquentDataTable
{
	return datatables()
		->eloquent($query)
		
		// Enable scout search engine for App\Models\Product
		->enableScoutSearchWith(Product::class)
		
		// Add filters for the scout search engine
		->scoutFilter(function ($keyword) {
			// ...
		})
		
		// Add filters that are applied after getting the results of the scout search engine / on the fallback search engine
		->filter(function ($query) {
			// ...
		}, true);
}
```

Process:
1. User types in search keyword
2. Check if scout search engine is enabled (`enableScoutSearchWith` method)
3. Apply scout search (or if it fails, fall back to default search)
    1. Raw search for the keyword + filters from `scoutFilter` method (limit: 1000 / all)
    2. Pluck the primary key of the results
    3. Apply a `whereIn('id', all_search_result_ids)` to the query
    4. Apply a `orderBy` to the query with the ordering of the search results
    5. Continue the normal process (adding the filters from `filter` method)

If there are any questions, feel free to hit me up :)

frknakk:
Fixes #3077 

@yajra
What do you think? Feel free to let me know if i should change some parts.
**disableOrdering js script**: We could either integrate that into laravel-datatables-html or add it into some docs for people to implement on their own (if they want to). I could also leave that part out completely. What do you think?

If it's possible to somehow add custom scss/js code into laravel-datatables-html i would also do 2 more pull requests for nice optimizations that i'm using since a long time on my own projects.

Also it would be good to add an example usage into docs:

```php
public function dataTable(QueryBuilder $query): EloquentDataTable
{
	return (new EloquentDataTable($query))
		// Enable scout search for eloquent model
		->enableScoutSearch(Product::class)
		
		// Add filters to scout search
		->scoutFilter(function (string $keyword) {
			return 'region IN ["Germany", "France"]'; // Meilisearch
                        return 'region:Germany OR region:France'; // Algolia
		})
		
		// Add filters to default search
		->filter(function (QueryBuilder $query, bool $scout_searched) {
			if (!$scout_searched)
			{
				// Filter already added for scout search
				$query->whereIn('region', ['Germany', 'France']);
			}

			// Stock is not indexed so it has to be filtered after the initial scout search
			$query->where('stock', '>', 50);
		}, true)

		->setRowId('id');
}
```

---

**Disable ordering (JS)**
When the scout search was performed, the code sets a `disableOrdering: true` flag in the ajax response. The JS part has to fetch that and disable the user ordering. Otherwise the user gets confused because the ordering is overwritten by the relevancy ordering of the search results.
```javascript
$('#products-table').on('xhr.dt', function (e, settings, json, xhr) {
	if (json == null || !('disableOrdering' in json)) return;

	let table = LaravelDataTables[$(this).attr('id')];
	if (json.disableOrdering) {
		table.settings()[0].aoColumns.forEach(function(column) {
			column.bSortable = false;
			$(column.nTh).removeClass('sorting_asc sorting_desc sorting').addClass('sorting_disabled');
		});
	} else {
		let changed = false;
		table.settings()[0].aoColumns.forEach(function(column) {
			if (column.bSortable) return;
			column.bSortable = true;
			changed = true;
		});
		if (changed) {
			table.draw();
		}
	}
});
```

### DIFF
--- a/src/QueryDataTable.php
+++ b/src/QueryDataTable.php
@@ -975,6 +975,12 @@ class QueryDataTable extends DataTableAbstract
         return $this;
     }
 
+    /**
+     * Apply scout search to query if enabled.
+     *
+     * @param  string  $search_keyword
+     * @return bool
+     */
     protected function applyScoutSearch(string $search_keyword): bool
     {
         if ($this->scoutModel == null) {
@@ -997,17 +1003,11 @@ class QueryDataTable extends DataTableAbstract
 
             // Order by scout search results & disable user ordering
             if (count($search_results) > 0) {
-                $escaped_ids = collect($search_results)
-                    ->map(function ($id) {
-                        return \DB::connection()->getPdo()->quote($id);
-                    })
-                    ->implode(',');
+                $this->applyFixedOrderingToQuery($this->scoutKey, $search_results);
 
-                $this->query->orderByRaw("FIELD($this->scoutKey, $escaped_ids)");
+                // Disable user ordering because we already order by search relevancy
+                $this->disableUserOrdering = true;
             }
-
-            // Disable user ordering because we already order by search relevancy
-            $this->disableUserOrdering = true;
 
             $this->scoutSearched = true;
 
@@ -1015,6 +1015,64 @@ class QueryDataTable extends DataTableAbstract
         } catch (\Exception) {
             // Scout search failed, fallback to default search
             return false;
+        }
+    }
+
+    /**
+     * Apply fixed ordering to query by a fixed set of values depending on database driver (used for scout search).
+     *
+     * Currently supported drivers: MySQL
+     *
+     * @param  string  $keyName
+     * @param  array  $orderedKeys
+     * @return void
+     * @throws \Exception If the database driver is unsupported.
+     */
+    protected function applyFixedOrderingToQuery(string $keyName, array $orderedKeys)
+    {
+        $connection = $this->query->getConnection();
+        $driver_name = $connection->getDriverName();
+
+        // Escape keyName and orderedKeys
+        $keyName = $connection->escape($keyName);
+        $orderedKeys = collect($orderedKeys)
+            ->map(function ($value) use ($connection) {
+                return $connection->escape($value);
+            });
+
+        switch ($driver_name)
+        {
+            case 'mysql':
+                // MySQL
+                $this->query->orderByRaw("FIELD($keyName, " . $orderedKeys->implode(',') . ")");
+                break;
+
+            /*
+            TODO: test implementations, fix if necessary and uncomment
+            case 'pgsql':
+                // PostgreSQL
+                $this->query->orderByRaw("array_position(ARRAY[" . $orderedKeys->implode(',') . "], $keyName)");
+                break;
+
+            case 'sqlite':
+            case 'sqlsrv':
+                // SQLite & Microsoft SQL Server
+
+                // should be generally compatible with all drivers using SQL syntax (but ugly solution)
+                $this->query->orderByRaw(
+                    "CASE $keyName "
+                    .
+                    $orderedKeys
+                        ->map(fn($value, $index) => "WHEN $keyName = $value THEN $index")
+                        ->implode(' ')
+                    .
+                    " END"
+                );
+                break;
+            */
+
+            default:
+                throw new \Exception("Unsupported database driver: $driver_name");
         }
     }
 
--- a/src/QueryDataTable.php
+++ b/src/QueryDataTable.php
@@ -1026,11 +1026,12 @@ class QueryDataTable extends DataTableAbstract
      * @param  string  $keyName
      * @param  array  $orderedKeys
      * @return void
+     *
      * @throws \Exception If the database driver is unsupported.
      */
     protected function applyFixedOrderingToQuery(string $keyName, array $orderedKeys)
     {
-        $connection = $this->query->getConnection();
+        $connection = $this->getConnection();
         $driver_name = $connection->getDriverName();
 
         // Escape keyName and orderedKeys
@@ -1040,36 +1041,35 @@ class QueryDataTable extends DataTableAbstract
                 return $connection->escape($value);
             });
 
-        switch ($driver_name)
-        {
+        switch ($driver_name) {
             case 'mysql':
                 // MySQL
-                $this->query->orderByRaw("FIELD($keyName, " . $orderedKeys->implode(',') . ")");
+                $this->query->orderByRaw("FIELD($keyName, ".$orderedKeys->implode(',').')');
                 break;
 
-            /*
-            TODO: test implementations, fix if necessary and uncomment
-            case 'pgsql':
-                // PostgreSQL
-                $this->query->orderByRaw("array_position(ARRAY[" . $orderedKeys->implode(',') . "], $keyName)");
-                break;
+                 /*
+                 TODO: test implementations, fix if necessary and uncomment
+                 case 'pgsql':
+                     // PostgreSQL
+                     $this->query->orderByRaw("array_position(ARRAY[" . $orderedKeys->implode(',') . "], $keyName)");
+                     break;
 
-            case 'sqlite':
-            case 'sqlsrv':
-                // SQLite & Microsoft SQL Server
+                 case 'sqlite':
+                 case 'sqlsrv':
+                     // SQLite & Microsoft SQL Server
 
-                // should be generally compatible with all drivers using SQL syntax (but ugly solution)
-                $this->query->orderByRaw(
-                    "CASE $keyName "
-                    .
-                    $orderedKeys
-                        ->map(fn($value, $index) => "WHEN $keyName = $value THEN $index")
-                        ->implode(' ')
-                    .
-                    " END"
-                );
-                break;
-            */
+                     // should be generally compatible with all drivers using SQL syntax (but ugly solution)
+                     $this->query->orderByRaw(
+                         "CASE $keyName "
+                         .
+                         $orderedKeys
+                             ->map(fn($value, $index) => "WHEN $keyName = $value THEN $index")
+                             ->implode(' ')
+                         .
+                         " END"
+                     );
+                     break;
+                 */
 
             default:
                 throw new \Exception("Unsupported database driver: $driver_name");
--- a/src/QueryDataTable.php
+++ b/src/QueryDataTable.php
@@ -1001,11 +1001,9 @@ class QueryDataTable extends DataTableAbstract
                 $this->query->whereIn($this->scoutKey, $search_results);
             });
 
-            // Order by scout search results & disable user ordering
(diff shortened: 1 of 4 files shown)


yajra (commented):
Was able to test and works well 🎉. However, it seems like the scout search is failing when the global search is off but it seems intended since meilisearch will basically do the wild search for us?

```php
		->filter(function (QueryBuilder $query, bool $scout_searched) {

		}, false)
```

frknakk:
> Was able to test and works well 🎉. However, it seems like the scout search is failing when the global search is off but it seems intended since meilisearch will basically do the wild search for us?
> 
> ```
> 		->filter(function (QueryBuilder $query, bool $scout_searched) {
> 
> 		}, false)
> ```

Yes, that's intended. Adding column search for scout would make the code a mess and difficult to maintain. Additionally i don't think there is a relevant use case for that. In 99% of the cases setting `searchableAttributes` on meilisearch/algolia will probably be enough.

yajra:
> Where do we put the disableOrdering js script?

I think we should not force disable the ordering function and let the developer handle it via javascript `ordering: false`.

frknakk:
> > Where do we put the disableOrdering js script?
> 
> I think we should not force disable the ordering function and let the developer handle it via javascript `ordering: false`.
> 
> In most of our cases, our users want a specific ordering but it will be a big boost in performance if we can delegate the searching via meilisearch.

I think you misunderstood the purpose of the script.

When searching via Meilisearch, it doesn't make any sense to keep the user ordering (e.g. sort by name asc), because then there will be dozens of unrelevant entries before you get the desired result. Because of that, when searching via Meilisearch the user ordering gets ignored serverside and it forces the order of meilisearch results (= entries that are matching best are shown first).

The disableOrdering js script is just a frontend script, that temporarily hides the ordering symbols when meilisearch was used. Because otherwise you can click on all columns to change ordering but nothing will change. The script also enables the ordering again when the search was deleted or meilisearch is not active.

yajra:
> When searching via Meilisearch, it doesn't make any sense to keep the user ordering (e.g. sort by name asc), because then there will be dozens of unrelevant entries before you get the desired result. Because of that, when searching via Meilisearch the user ordering gets ignored serverside and it forces the order of meilisearch results (= entries that are matching best are shown first).

Good point, I was thinking only of search results (discarding relevance) but with ordering as defined by the users.

> The disableOrdering js script is just a frontend script, that temporarily hides the ordering symbols when meilisearch was used. Because otherwise you can click on all columns to change ordering but nothing will change. The script also enables the ordering again when the search was deleted or meilisearch is not active.

I was trying to do a script for this and can't figure it out ^_^. Do you have a sample script for this?

frknakk:
> I was trying to do a script for this and can't figure it out ^_^. Do you have a sample script for this?

Yes, it's above in my initial message 😄

yajra:
OMG! Apologies for overlooking that script 😅. The functionality makes full sense to me now.

I think it would be nice if we could have that script integrated into the HTML builder but with docs too for those not using the builder. Maybe we can add new script template for scout to integrate this?


**scout.blade.php**

```js
$(function(){
    window.{{ config('datatables-html.namespace', 'LaravelDataTables') }}=window.{{ config('datatables-html.namespace', 'LaravelDataTables') }}||{};
    window.{{ config('datatables-html.namespace', 'LaravelDataTables') }}["%1$s"]=$("#%1$s").DataTable(%2$s);

    $('#%1$s').on('xhr.dt', function (e, settings, json, xhr) {
        if (json == null || !('disableOrdering' in json)) return;

        let table = LaravelDataTables[$(this).attr('id')];
        if (json.disableOrdering) {
            table.settings()[0].aoColumns.forEach(function(column) {
                column.bSortable = false;
                $(column.nTh).removeClass('sorting_asc sorting_desc sorting').addClass('sorting_disabled');
            });
        } else {
            let changed = false;
            table.settings()[0].aoColumns.forEach(function(column) {
                if (column.bSortable) return;
                column.bSortable = true;
                changed = true;
            });
            if (changed) {
                table.draw();
            }
        }
    });
});
```

And then maybe a shortcut builder method to set the template to scout?

```php
->setTableId('products-table')
->template('scout');
```

src/QueryDataTable.php
```diff
@@ -861,4 +920,154 @@ public function ignoreSelectsInCountQuery(): static
...
+
+            // Order by scout search results & disable user ordering
+            if (count($search_results) > 0) {
+                $escaped_ids = collect($search_results)
+                    ->map(function ($id) {
+                        return \DB::connection()->getPdo()->quote($id);
+                    })
+                    ->implode(',');
+
+                $this->query->orderByRaw("FIELD($this->scoutKey, $escaped_ids)");
```
yajra, line 1006: This might break on other database drivers.

src/QueryDataTable.php
```diff
@@ -861,4 +920,154 @@ public function ignoreSelectsInCountQuery(): static
...
+            // Apply scout search results to query
+            $this->query->where(function ($query) use ($search_results) {
+                $this->query->whereIn($this->scoutKey, $search_results);
+            });
+
+            // Order by scout search results & disable user ordering
+            if (count($search_results) > 0) {
+                $escaped_ids = collect($search_results)
+                    ->map(function ($id) {
+                        return \DB::connection()->getPdo()->quote($id);
```
yajra, line 1002: I think we should get the connection from the scout model instance.

yajra (changes requested):
I also created a [poc repo](https://github.com/yajra/poc-dt-scout) to test this. Thanks!

src/QueryDataTable.php
yajra: Maybe we can extract this to a method and throw an exception if a DB is not yet supported? We can do matching via DB driver name `$model->getConnection()->getDriverName()`.

My team uses Oracle DB and would love to add support to this in future releases.

frknakk: Done! :)

src/QueryDataTable.php
frknakk: Done! :)

src/QueryDataTable.php
frknakk: Done! :)

src/QueryDataTable.php
frknakk: Done! :)

frknakk:
@yajra Do you have any more suggestions for improvement? Looks good to me :)

Last thing missing is a docs section that has a quick start with the usage examples from both this pull request and the new one in laravel-datatables-html.

yajra (approved):

yajra:
The changes look good! Will test on other DB drivers.

src/QueryDataTable.php
```diff
@@ -861,4 +920,212 @@ public function ignoreSelectsInCountQuery(): static
...
+        return $this;
+    }
+
+    /**
+     * Apply scout search to query if enabled.
+     *
+     * @param  string  $search_keyword
+     * @return bool
+     */
+    protected function applyScoutSearch(string $search_keyword): bool
```
yajra, line 984: I was testing against SQLite and it seems like the `scoutSearch` flag is not being set to true due to the exception on fixed ordering.

**Suggestion:**

```php
    protected function applyScoutSearch(string $search_keyword): bool
    {
        if ($this->scoutModel == null) {
            return false;
        }

        try {
            // Perform scout search
            $search_filters = '';
            if (is_callable($this->scoutFilterCallback)) {
                $search_filters = ($this->scoutFilterCallback)($search_keyword);
            }

            $search_results = $this->performScoutSearch($search_keyword, $search_filters);

            // Apply scout search results to query
            $this->query->where(function ($query) use ($search_results) {
                $this->query->whereIn($this->scoutKey, $search_results);
            });

            $this->scoutSearched = true;

            // Order by scout search results & disable user ordering
            if (count($search_results) > 0) {
                $this->applyFixedOrderingToQuery($this->scoutKey, $search_results);

                // Disable user ordering because we already order by search relevancy
                $this->disableUserOrdering = true;
            }

            return true;
        } catch (\Exception) {
            // Scout search failed, fallback to default search
            return false;
        }
    }
```

frknakk: Replaced exception with bool response, so that the system falls back to user ordering if fixed ordering by relevancy is not supported. Also fixed and enabled sql implementation (tested sqlite but its general sql syntax so it should work on mssql too)

yajra:
Confirmed working on SQLite. The CASE WHEN SQL doesn't seem to work with Oracle but it's ok.

Will also try PGSQL by tomorrow.

yajra:
PostgreSQL does not work. However was able to do a code that works for both pgsql and oracle. Can't test on `sqlsrv` though.

```php
    protected function applyFixedOrderingToQuery(string $keyName, array $orderedKeys)
    {
        $connection = $this->getConnection();
        $driverName = $connection->getDriverName();

        // Escape keyName and orderedKeys
        $keyName = $connection->getQueryGrammar()->wrap($keyName);
        $orderedKeys = collect($orderedKeys)
            ->map(function ($value) use ($connection) {
                return $connection->escape($value);
            });

        switch ($driverName) {
            case 'mysql':
                $this->query->orderByRaw("FIELD($keyName, ".$orderedKeys->implode(',').')');
                return true;

            case 'pgsql':
            case 'oracle':
                $this->query->orderByRaw(
                    "CASE "
                    .
                    $orderedKeys
                        ->map(fn($value, $index) => "WHEN $keyName=$value THEN $index")
                        ->implode(' ')
                    .
                    " END"
                );
                return true;

            case 'sqlite':
            case 'sqlsrv':
                $this->query->orderByRaw(
                    "CASE $keyName "
                    .
                    $orderedKeys
                        ->map(fn($value, $index) => "WHEN $value THEN $index")
                        ->implode(' ')
                    .
                    " END"
                );
                return true;

            default:
                return false;
        }
    }
```

Merging this and will apply the patch in a new PR. Thanks a lot for your patience and contribution! 🍻 🎉

yajra:
Released on v10.11.0 🚀 Thanks!

PR #3098, opened 2023-11-25
fix: Prevent error when PHP extension iconv not enabled.

sangnguyenplus:
This PR will fix error when PHP extension `iconv` not enabled, so we can't use function `iconv()`.

```
Call to undefined function Yajra\DataTables\iconv()
 {"userId":1,"exception":"[object] (Error(code: 0): Call to undefined function Yajra\\DataTables\\iconv() 
at .../vendor/yajra/laravel-datatables-oracle/src/QueryDataTable.php:867)
```

### DIFF
--- a/src/QueryDataTable.php
+++ b/src/QueryDataTable.php
@@ -943,6 +943,7 @@ class QueryDataTable extends DataTableAbstract
      * @param  string  $model
      * @param  int  $max_hits
      * @return $this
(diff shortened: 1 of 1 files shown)


yajra (approved):

yajra:
Released on v10.11.1 🚀 Thanks!

PR #3105, opened 2023-12-11
fix: scout search with smart search

frknakk:
Hi !

Here is a optimization for scout search (if smart search is enabled).

Old: keyword is exploded using spaces into multiple keywords and then each keyword gets scout-searched
New: whole keyword gets scout-searched (without exploding), the search engine deals with multiple words


yajra:
Released on v10.11.2 🚀 Thanks!

PR #3107, opened 2023-12-25
fix: Update composer.json to use Larastan Org

arnebr:
Starting with Larastan 2.7.0, the Larastan repository will now be managed under the Larastan organization.


yajra:
Released on v10.11.3 🚀 Thanks!

PR #3135, opened 2024-04-12
feat: Optimize simple queries

Arkhas:
The issues was for simple queries with complexe wheres, a sub query was used resulting in the following query : 

```sql
select count(*)
from (select "users".*
      from "users"
      where (select "title"
             from "posts"
             where "posts"."user_id" = "users"."id"
             order by "created_at" asc
    ) = ?
) count_row_table
```
resulting into an uneccessary temporary table, the query we would like to have would be this one : 

```sql
select count(*)
from "users"
where (select "title"
       from "posts"
       where "posts"."user_id" = "users"."id"
       order by "created_at" asc
) = ?
```

On our app, the gains were massive, one of our datables would just timeout and with this change it load now in less than 5 seconds

### DIFF
--- a/src/QueryDataTable.php
+++ b/src/QueryDataTable.php
@@ -176,7 +176,12 @@ class QueryDataTable extends DataTableAbstract
             $builder->select(DB::raw('1 as dt_row_count'));
             $clone = $builder->clone();
             $clone->setBindings([]);
-            $clone->wheres = [];
+            if ($clone instanceof EloquentBuilder) {
+                $clone->getQuery()->wheres = [];
+            } else {
+                $clone->wheres = [];
+            }
+
             if ($this->isComplexQuery($clone)) {
                 if (!$this->ignoreSelectInCountQuery) {
                     $builder = clone $this->query;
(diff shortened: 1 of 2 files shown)


yajra:
I guess this PR also fixes #3133. @abdulrhmansouda would you be able to review? Thanks!

abdulrhmansouda:
I fixed this problem when I changed the isComplexQuery function #3133.
and it's ready. you can review it.

yajra:
Please run `composer pr` on your local machine to fix the linting. I haven't figured out yet how to automate in PR. Thanks!

src/QueryDataTable.php
```diff
@@ -174,34 +174,40 @@ public function prepareCountQuery(): QueryBuilder
...
-                ->setBindings($builder->getBindings());
+                return $this->getConnection()
+                            ->query()
+                            ->fromRaw('(' . $builder->toSql() . ') count_row_table')
+                            ->setBindings($builder->getBindings());
+            }
         }
-
         $row_count = $this->wrap('row_count');
         $builder->select($this->getConnection()->raw("'1' as {$row_count}"));
```
yajra, line 197: Maybe we can consider @abdulrhmansouda suggestion of removing the orders?

```suggestion
        $builder->select($this->getConnection()->raw("'1' as {$row_count}"))->reorder();
```

yajra (changes requested):
The changes look good so far. 

Please fix the code style using `composer pint` or `composer pr` to run all quality tools. It will make the PR easier to review.

Thanks!

yajra (approved):

yajra:
Released on v11.1.0, 🚀 thanks!

PR #3146, opened 2024-05-23
fix: mariadb support for scout search

frknakk:
Laravel 11 introduced a mariadb database driver. We can use the same code as for mysql.


yajra:
Released https://github.com/yajra/laravel-datatables/releases/tag/v11.1.1 :rocket: thanks!

PR #3154, opened 2024-07-02
fix: ErrorException when direction is null

mgralikowski:
In most cases, the `dir` attribute is empty (null) which causes a notice error in the recent PHP version.

> ErrorException: strtolower(): Passing null to parameter #1 ($string) of type string is deprecated in /var/www/html/vendor/yajra/laravel-datatables-oracle/src/Utilities/Request.php:101

### DIFF
--- a/src/Utilities/Request.php
+++ b/src/Utilities/Request.php
@@ -86,7 +86,7 @@ class Request
             /** @var string $direction */
             $direction = $this->request->input("order.$i.dir");
 
-            $order_dir = strtolower($direction) === 'asc' ? 'asc' : 'desc';
+            $order_dir = $direction && strtolower($direction) === 'asc' ? 'asc' : 'desc';
             if ($this->isColumnOrderable($order_col)) {
                 $orderable[] = ['column' => $order_col, 'direction' => $order_dir];
             }


src/Utilities/Request.php
```diff
@@ -86,7 +86,7 @@ public function orderableColumns(): array
             /** @var string $direction */
             $direction = $this->request->input("order.$i.dir");
```
yajra, line 87: Good catch @mgralikowski, maybe we can just set the default value here instead?

```suggestion
            $direction = $this->request->input("order.$i.dir", 'desc');
```

yajra: I think it's ok since ensuring 'asc' or 'desc' is a security protection for SQL injection attacks.

Or we can fix the docblock instead to `/** @var string|null $direction */`

mgralikowski: Docblock would not resolve the PHP error ;)

IMHO - my proposal is more clear, because in so, there is only a single place where you define a default sorting - `desc`, no matter whether the passed value is undefined (null), incorrect, or is an attempt of SQL injection  ;) 

Another suggestion if you don't like this one is cast input to string:

`(string) $this->request->input("order.$i.dir");`

yajra: > Docblock would not resolve the PHP error ;)

Oh, I was thinking about the static analysis since it's currently set as a string but the value can be null which is why you came up with the proposal to check if the `$direction` is set. ^_^

yajra: Anyway, this is good to go.

yajra:
Released on https://github.com/yajra/laravel-datatables/releases/tag/v11.1.2, thanks! 🚀

PR #3157, opened 2024-07-09
fix: make query for filteredRecords when totalRecords was manually set

JurianArie:
Hi, we always use `setTotalRecords(0)` since we only show the filtered records count. After upgrading to Laravel 11 we noticed that our pagination stopped working. This seems to be caused by c37494f11f0b6e198fa7428edd38e5d5b2a6d9de. This PR changes the logic to always query the filtered count if the total count was set manually.

### DIFF
--- a/tests/Integration/QueryDataTableTest.php
+++ b/tests/Integration/QueryDataTableTest.php
@@ -26,7 +26,7 @@ class QueryDataTableTest extends TestCase
         $crawler->assertJson([
             'draw' => 0,
             'recordsTotal' => 10,
-            'recordsFiltered' => 10,
+            'recordsFiltered' => 20,
         ]);
     }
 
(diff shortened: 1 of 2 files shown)


tests/Integration/QueryDataTableTest.php
JurianArie, line 1: The changes from c37494f11f0b6e198fa7428edd38e5d5b2a6d9de to the tests have been reverted, the optimalization should not have changed this behavior.

yajra:
Thank you for the PR, I will review this as soon as I can.

yajra:
@Arkhas it seems like the PR you've sent is being reverted on this change. Would you mind reviewing and approving this? Thanks!

yajra:
I confirmed one less query after the patch. This is a good optimization trick for large datasets. Thanks!

Released on [11.1.3](https://github.com/yajra/laravel-datatables/releases/tag/v11.1.3) :rocket:

PR #3169, opened 2024-08-20
Prevent unwanted count query

OzanKurt:
The only reason we're using `setTotalRecords` is to prevent the count query.

Even though we're seting the `totalRecords` we also need to set `performedTotalRecordsCount` so that;

IF -> there are filters applied to the query (any kind of search)
    THEN -> don't perform total count **AGAIN**

There was an update about this topic recently: https://github.com/yajra/laravel-datatables/commit/9f7169a1cea51ff22cf4154c5c0e6da35d19c558

Sadly it didn't cover all the cases.

Here is the related code in [QueryDataTable#L275](https://github.com/yajra/laravel-datatables/blob/9f7169a1cea51ff22cf4154c5c0e6da35d19c558/src/QueryDataTable.php#L275)

```php
// If no modification between the original query and the filtered one has been made
// the filteredRecords equals the totalRecords
if ($this->query == $initialQuery && $this->performedTotalRecordsCount) {
    $this->filteredRecords ??= $this->totalRecords;
} else {
    $this->filteredCount();
}
```

### DIFF
--- a/src/DataTableAbstract.php
+++ b/src/DataTableAbstract.php
@@ -524,7 +524,6 @@ abstract class DataTableAbstract implements DataTable
     public function setTotalRecords(int $total): static
     {
         $this->totalRecords = $total;
-        $this->performedTotalRecordsCount = true;
 
         return $this;
     }
--- a/src/DataTableAbstract.php
+++ b/src/DataTableAbstract.php
@@ -524,6 +524,7 @@ abstract class DataTableAbstract implements DataTable
     public function setTotalRecords(int $total): static
     {
         $this->totalRecords = $total;
+        $this->performedTotalRecordsCount = true;
 
         return $this;
     }
--- a/src/QueryDataTable.php
+++ b/src/QueryDataTable.php
@@ -161,6 +161,19 @@ class QueryDataTable extends DataTableAbstract
 
         return $this;
     }
+    
+    /**
+     * Set total records manually.
+     *
+     * @return $this
+     */
+    public function setTotalRecords(int $total): static
+    {
+        parent::setTotalRecords($total);
+        $this->performedTotalRecordsCount = true;
+
+        return $this;
+    }
 
     /**
      * Count total items.
--- a/src/QueryDataTable.php
+++ b/src/QueryDataTable.php
@@ -161,7 +161,7 @@ class QueryDataTable extends DataTableAbstract
 
         return $this;
     }
-    
+
     /**
      * Set total records manually.
      *
--- a/tests/Integration/QueryDataTableTest.php
+++ b/tests/Integration/QueryDataTableTest.php
@@ -26,7 +26,7 @@ class QueryDataTableTest extends TestCase
         $crawler->assertJson([
             'draw' => 0,
             'recordsTotal' => 10,
-            'recordsFiltered' => 20,
+            'recordsFiltered' => 10,
         ]);
     }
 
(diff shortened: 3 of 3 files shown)


OzanKurt:
Can we please merge this asap? @yajra

yajra:
The changes look good but ci/cd checks are failing. Can you please fix it? Thanks!

OzanKurt:
As far as I see it's only 1 test failing am I right?

image

I will update the test so that it matches the expected behaviour.

yajra:
It seems like 2 tests failed:

https://github.com/yajra/laravel-datatables/actions/runs/10478462086/job/29021885548?pr=3169#step:7:21

https://github.com/yajra/laravel-datatables/actions/runs/10478462086/job/29021885548?pr=3169#step:7:273

JurianArie:
This breaks my use case wherein we only want to make a query for the filtered count but never for the total count.

OzanKurt:
> This breaks my use case wherein we only want to make a query for the filtered count but never for the total count.

Are you sure? This is exactly what I am intending as well.

OzanKurt:
Finally fixed all the checks 😄

JurianArie:
I tested the changes locally, and I can confirm that it breaks my use case.

> `if ($this->query == $initialQuery ...)` part check if there is a modification to the "filtered" query. Which should result in `false && true` which will call `$this->filteredCount();`

This assumes that filters are applied. If no filters are applied, it will result in `true && true`.

OzanKurt:
Can you please send your DataTable so that I can test as well?

JurianArie:
The test cases before reflected the correct behavior. Here, the users table contains 20 records, and we manually set the total records to zero. The filtered records should still be 20.

https://github.com/yajra/laravel-datatables/blob/178a25d9668ac822bc9b57f2dc1ea3abfb66ca96/tests/Integration/QueryDataTableTest.php#L462-L464

https://github.com/yajra/laravel-datatables/blob/178a25d9668ac822bc9b57f2dc1ea3abfb66ca96/tests/Integration/QueryDataTableTest.php#L34-L42

OzanKurt:
Exactly what I mean.

Having `0` total records and displaying `20` of them doesn't make any sense.

JurianArie:
You should think of `setTotalRecords(0)` as being the same as calling the now deprecated `skipTotalRecords`. I want to skip the total records count, not the filtered records count.

OzanKurt:
### Case 1: (Your case)

```php
// Total 2500: / Filtered: 100
$dataTable->setTotalRows(0)->make();
```

This should result in:
```
    No Filter -> "Displaying 100 of 0 rows"
    Filtered  -> "Displaying 100 of 0 rows"
```

### Case 2: (My case)

```php
// Total 2500: / Filtered: 100
$dataTable->setTotalRows(2500)->make();
```

This should result in:
```
    No Filter -> "Displaying 2500 of 2500 rows"
    Filtered  -> "Displaying 100 of 2500 rows"
```

Am I correct about this?

JurianArie:
<details>
<summary>1. Fresh setup of using the quick start guide</summary>

Guide: https://yajrabox.com/docs/laravel-datatables/master/quick-starter
image


</details>

<details>
<summary>2. Update UsersDataTable::dataTable with setTotalRecords(0)</summary>

image


</details>

</details>

<details>
<summary>3. Apply your change</summary>

image


</details>

As you can see when I apply your change the pagination is broken.

OzanKurt:
I still can't understand the logic.

My point is:

**There is no reason to query the filtered count, IF the query is not being filtered.**

### Your version:

image
image

In both instances you will send a count query.

### My version:

image
image

yajra:
Fixed via #3170

PR #3171, opened 2024-08-26
Set maximum records limit per-page in config.

gemul:
This PR adds a `max_limit` configuration to prevent end-users from arbitrarily setting a large value for the `length` request parameter, which could otherwise cause the server to overload if not handled manually (e.g., in middleware).

If the `length` value exceeds this limit, it will be overridden with the specified `max_limit`. 

The default value is 100.


yajra:
Thanks for the PR! The changes look good, however, I suspect this might break the export functionality of [buttons package](https://github.com/yajra/laravel-datatables-buttons).

Will check further when I get the chance.

Issue #2871, opened 2022-10-07
Can't use column searches in same table relations


affaron:
Hi!

Recently I discovered an issue when using multiple column search with different relation using the same table. Here's what I mean.

I have a model with relations:
```php
public function country_from()
    {
        return $this->belongsTo(Country::class, 'from_country_id');
    }

    public function country_to()
    {
        return $this->belongsTo(Country::class, 'to_country_id');
    }
```
and two columns in the datatable with these relation (country_from.name, country_to.name). I need to use filtering by both columns at the same time and I got a problem: when the second relation is applied the join is omitted since it's already joined and same table used in WHERE condition: 
```sql
select
  `borders`.*
from
  `borders`
  left join `countries` on `borders`.`from_country_id` = `countries`.`id`
where
  LOWER(`countries`.`name`) LIKE '%south africa%'
  and LOWER(`countries`.`name`) LIKE '%zimbabwe%'
```
This happens because of this code:
https://github.com/yajra/laravel-datatables/blob/9e96d1f4151ef33d7046f5581280730d98fb2d1d/src/EloquentDataTable.php#L251-L253

It makes since until you need to join the same table. I figured out a way to make it work like it supposed to in my case, but Im not sure it's PR worth hack and it's not thoroughly tested :) Here's what I did. 
Generate an alias for the table if it's already joined and return it with performJoin method (sorry, not the cleanest code):
```php
protected function performJoin($table, $foreign, $other, $type = 'left')
    {
        $alias = $table;
        $joins = [];

        foreach ((array) $this->getBaseQueryBuilder()->joins as $key => $join) {
            $joins[] = Str::before($join->table, ' as ');
        }

        if (in_array($table, $joins)) {
            $index = count(array_filter($joins, function ($n) use ($table) { return $n === $table; })) + 1;
            $alias = $table . '_' . $index;
            $other = str_replace($table, $alias, $other);
            $table = $table . ' as ' . $alias;
        }

        $this->getBaseQueryBuilder()->join($table, $foreign, '=', $other, $type);

        return $alias;
    }
```
and use this alias in the end of `joinEagerLoadedColumn()` method:
```php
            $alias = $this->performJoin($table, $foreign, $other);
            $lastQuery = $model->getQuery();
        }

        return $alias.'.'.$relationColumn;
```

after that the lib can generate accurate query:
```sql
select
  `borders`.*
from
  `borders`
  left join `countries` on `borders`.`from_country_id` = `countries`.`id`
[...]

PR #3216, opened 2025-02-21
fix: prevent duplicate table name errors

Seb33300:
Fixes https://github.com/yajra/laravel-datatables/issues/2871/

This PR fixes the duplicate table name SQL error by adding aliases to joined tables.
It also supports multi levels joins:
Eg:
```
- article.author.name
- article.category.created_by.name
- article.comments[].user.name
```


Seb33300:
Hi @yajra, PR is now ready for review 🙏

yajra:
@Seb33300 can you provide an example of the table aliases generated in this PR? One concern I have is the Oracle limitation that only accepts up to 30 chars for object/table name length.

Initial Laravel 12 support merged on the master branch.

Thanks!

Seb33300:
@yajra It concatenates all parent relation names to create a unique alias name:
```
updated_by.name => _updated_by.name
post.created_by.name => _post_created_by.name
```

Here is an example of query generated:
```sql
select `users`.* from `users` left join `users` as `_updated_by` on `users`.`updated_user_id` = `_updated_by`.`id` where `_updated_by`.`name` LIKE '%%seb%%' order by `_updated_by`.`name` asc limit 10 offset 0
```

Seb33300:
> One concern I have is the Oracle limitation that only accepts up to 30 chars for object/table name length.

Please note that starting Oracle Database 12c Release 2 (12.2) released in 2017, the maximum size is now 128 chars.
https://oracle-base.com/articles/12c/long-identifiers-12cr2

Seb33300:
Unless you have columns with multiple joins, It's unlikely to happen, but still theoretically possible.

I workaround I can suggest is to use md5 to generate a hash of the alias in case it's more than 30 chars.

yajra:
Thanks and will review first on an actual project. I just recalled that we rarely use relation and opt to use join statements for more stable results.

yajra:
Tested on a new Laravel 12 app and works as expected. Tested self-join SQL and seems to be fixed:

SQL Generated:

```sql
select * from "users" left join "users" as "_creator" on "users"."created_by" = "_creator"."id" order by "_creator"."name" desc limit 10 offset 0
```

Is the `_` on alias (`_creator`) intended?

Seb33300:
> ~~Is the `_` on alias (`_creator`) intended?~~

Yes, this is to prevent a potential conflict with a table name if the relation name is the same as the table name.

yajra:
Thanks for the feedback. Released on v12 🚀

Arkhas:
> ### ⚠️ This PR includes a BC break
> As you can see with the test case I modified, this PR comes with a small BC break. If a column has defined a custom `order by` using `orderColumn()`, the table is now using an alias that will not match with the table name. A new parameter has been added to the `orderColumn()` callable to pass the column name with alias.
> 
> ```
> DataTables::eloquent($model)
>     ->orderColumn('relation.name', function ($query, $order, $column) {
>         // $column = alias.name of the relation.name column
>         $query->orderBy($column, $order);
>     });
> ```

That's not a small breaking change, every time we do a custom order we have to modify our code for example (not real code) : 

```
DataTables::eloquent($model)
     ->orderColumn('item.total', function ($query, $order) {
         $query->orderBy(DB::raw('cart_items.amount * cart_items.quantity'), $order);
     });
```

we have to change it to : 

```
DataTables::eloquent($model)
     ->orderColumn('item.total', function ($query, $order, $column) {
     	 $alias = explode('.', $column)[0];
         $query->orderBy(DB::raw("$alias.amount * $alias.quantity"), $order);
     });
```

please, find another way of doing it ? 

Plus, why keeping the leading `_` ? juste remove it if first char ?

Seb33300:
> > ~Is the `_` on alias (`_creator`) intended?~
> 
> Yes, this is to prevent a potential conflict with a table name if the relation name is the same as the table name.

Hi @Arkhas, yes the `_` is intended to prevent conflict, I need to do some additional test to confirm if we can remove it or not.

Also in https://github.com/yajra/laravel-datatables/pull/3229 we added a resolver to the `editColumn()`.
Maybe we can do the same with `orderColumn()` to simplify things.

Arkhas:
having the column and the table as 2 seperate parameters would help : 

```
->orderColumn('user.id', function ($query, $order, $column, $tableAlias)
```

but maybe we can have it as a configuration option ? we are struggling updating the package because of this commit :/

Seb33300:
@Arkhas I finally had time to take a look on this and it's unfortunately not possible to as backward compatibility.
Even if we remove the front `_` it will not help because we use the relation name, not the table name.
For example `user.name` will result in `JOIN users as _user`.
This is to support cases where we can join the same table multiple times such as:
- `created_by.name` => `JOIN users as _created_by`
- `updated_by.name` => `JOIN  users as _updated_by`
Before, this was not working and resulted in an error.

In order to help, I created a [new PR](https://github.com/yajra/laravel-datatables/pull/3232) to replace the `$column` parameter by a `$resolver` parameter.
The resolver is callable you can use to get the alias from a column, in you case:

```php
DataTables::eloquent($model)
     ->orderColumn('item.total', function ($query, $order, $resolver) {
     	 $alias = explode('.', $column)[0];
         $query->orderBy(DB::raw("{$resolver('item.amount ')} * {$resolver('item.quantity')}"), $order);
     });
```

Arkhas:
that would not help a lot, we would just need the table alias, (the name of the column is totally irrelevant as we are using a custom ordering.

what would be nice: 

```
->orderColumn('item.total', function ($query, $order, $tableAlias) {
```

but what we would need is to be able to not use this feature to not completely break our app like something like this : 

```
			if(config('datatables.use_smart_alias'){
				$lastAlias = $tableAlias ?: $lastQuery->getModel()->getTable();
	            $tableAlias = $tableAlias.'_'.$eachRelation;
	            $pivotAlias = $tableAlias.'_pivot';
            }else{
				$pivotAlias = $model->getTable();
			}
```

or something like that

Seb33300:
The `$tableAlias` param will work only if you work with a single table.
But in more complex cases, if you need to work on values coming from different tables, this will not work.

for example if you have a case where you need to get the tax rate in another table, using the resolver you can do:
```php
$resolver('item.amount') * $resolver('item.tax_rule.rate')
```

The resolver can be used to resolve relations between models.

I can try to implement a BC option as you suggested, but if that option is not enabled, you may face ambiguous column name errors.

Seb33300:
Another option you may use is to re create the join by yoursel.
On you base query, you can just add 
```php
$query->join('cart_items', ...)
```

And then, you will be able to use the table name `cart_items` like before.

Arkhas:
>I can try to implement a BC option as you suggested, but if that option is not enabled, you may face ambiguous column name errors.

it would be easier for us to upgrade, maybe on a datatable basis instead of a global configuration so that we could upgrade slowly each of our Datatables : 

```
$datatable = new EloquentDataTable($this->query());
$datatable->smartTableAlias(false);
```

something like that ? maybe more versatile

Seb33300:
Hi @Arkhas, please check that PR: https://github.com/yajra/laravel-datatables/pull/3234

Arkhas:
>Hi @Arkhas, please check that PR: https://github.com/yajra/laravel-datatables/pull/3234

Sounds good, maybe add the ability to set a default value ? that's would be nice to have, but we can work with this if you don't think that would be a good idea.

Thank you very much !

PR #3224, opened 2025-04-03
fix: misc improvements

Seb33300:
I created a custom action where I need to return a json with complex manipulation on the data.
In that action I need to access the query results.
This is an improvement to prevent the query to be triggered twice.

### DIFF
--- a/src/DataTableAbstract.php
+++ b/src/DataTableAbstract.php
@@ -779,7 +779,7 @@ abstract class DataTableAbstract implements DataTable
     /**
      * Count filtered items.
      */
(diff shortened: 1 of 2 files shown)


src/DataTableAbstract.php
```diff
@@ -779,7 +779,7 @@ protected function searchPanesSearch(): void
     /**
      * Count filtered items.
      */
-    protected function filteredCount(): int
+    public function filteredCount(): int
```
Seb33300, line 782: I modified the visibility of that method to public because the `totalCount()` method is also public.

PR #3227, opened 2025-04-22
fix: prevent ambiguous column names

Seb33300:
When using complex queries with joins, I am often facing ambiguous column name errors when sorting a column of the base table sharing the same name as a column of a relation table field (eg: created_at)

Until now, the workaround I used was to rename my columns
``` php
// From
DateColumn::make('created_at')`
// To
DateColumn::make('created_at', 'table.created_at')
```
But it's probably better to fix the root cause by forcing the table name if the column is not a relation.

And it's consistent with `$this->joinEagerLoadedColumn()` which is returning columns name with alias.

### DIFF
--- a/src/EloquentDataTable.php
+++ b/src/EloquentDataTable.php
@@ -167,7 +167,7 @@ class EloquentDataTable extends QueryDataTable
         $relation = str_replace('[]', '', implode('.', $parts));
 
         if ($this->isNotEagerLoaded($relation)) {
-            return $column;
+            return parent::resolveRelationColumn($column);
         }
 
         return $this->joinEagerLoadedColumn($relation, $columnName);
@@ -188,14 +188,14 @@ class EloquentDataTable extends QueryDataTable
         $lastQuery = $this->query;
         foreach (explode('.', $relation) as $eachRelation) {
             $model = $lastQuery->getRelation($eachRelation);
-            $lastAlias = $tableAlias ?: $lastQuery->getModel()->getTable();
+            $lastAlias = $tableAlias ?: $this->getTablePrefix($lastQuery);
             $tableAlias = $tableAlias.'_'.$eachRelation;
             $pivotAlias = $tableAlias.'_pivot';
             switch (true) {
                 case $model instanceof BelongsToMany:
                     $pivot = $model->getTable().' as '.$pivotAlias;
                     $pivotPK = $pivotAlias.'.'.$model->getForeignPivotKeyName();
-                    $pivotFK = $lastAlias.'.'.$model->getParentKeyName();
+                    $pivotFK = ltrim($lastAlias.'.'.$model->getParentKeyName(), '.');
                     $this->performJoin($pivot, $pivotPK, $pivotFK);
 
                     $related = $model->getRelated();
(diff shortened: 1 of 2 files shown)


src/EloquentDataTable.php
```diff
@@ -166,7 +166,9 @@ protected function resolveRelationColumn(string $column): string
         $columnName = array_pop($parts);
         $relation = str_replace('[]', '', implode('.', $parts));
 
-        if ($this->isNotEagerLoaded($relation)) {
+        if (! $relation) {
+            return $this->query->getModel()->getTable().'.'.$column;
```
yajra, line 170: I think this will fail in cases where we use a table alias?

Seb33300: Good catch, I will try to check this

Seb33300: Should be fixed now.
I just noticed a `addTablePrefix()` method was already created for this purpose.
So I just reused it.

src/QueryDataTable.php
```diff
@@ -471,19 +471,32 @@ protected function compileQuerySearch($query, string $column, string $keyword, s
...
+    }
 
-                $column = $from.'.'.$column;
+    /**
+     * Try to get the base table prefix.
+     * To be used to prevent ambiguous field name.
+     *
+     * @param  QueryBuilder|EloquentBuilder  $query
+     */
+    protected function getTablePrefix($query): ?string
```
Seb33300, line 486: I extracted the code previously in `addTablePrefix()` to a separated method

yajra:
@Seb33300 the patch looks great. However, found a failing use case:

Given this query:

```php
    public function query(User $model): QueryBuilder
    {
        return $model->newQuery()
            ->withCount('permissions');
    }
```

With column defined as:

```php
            Column::computed('permissions_count')
                ->title('Permissions')
                ->orderable()
```

The expectation is the ordering should still work. Since the patch adds a prefix, the ordering now fails.


```sql
Exception Message:

Error Code    : 904
Error Message : ORA-00904: "USERS"."PERMISSIONS_COUNT": invalid identifier
Help: https://docs.oracle.com/error-help/db/ora-00904/
Position      : 289
Statement     : select t2.* from ( select rownum AS "rn", t1.* from (select "USERS".*, (select count(*) from "PERMISSIONS" inner join "PERMISSION_USER" on "PERMISSIONS"."ID" = "PERMISSION_USER"."PERMISSION_ID" where "USERS"."ID" = "PERMISSION_USER"."USER_ID") as "PERMISSIONS_COUNT" from "USERS" order by "USERS"."PERMISSIONS_COUNT" asc) t1 ) t2 where t2."rn" between 1 and 20
Bindings      : []
 (Connection: oracle, SQL: select t2.* from ( select rownum AS "rn", t1.* from (select "USERS".*, (select count(*) from "PERMISSIONS" inner join "PERMISSION_USER" on "PERMISSIONS"."ID" = "PERMISSION_USER"."PERMISSION_ID" where "USERS"."ID" = "PERMISSION_USER"."USER_ID") as "PERMISSIONS_COUNT" from "USERS" order by "USERS"."PERMISSIONS_COUNT" asc) t1 ) t2 where t2."rn" between 1 and 20)
```

Seb33300:
@yajra I just pushed a fix for this

yajra:
Wow, this patch is awesome @Seb33300 💯! This will solve a lot of previous issues. 

Tested on several unique use cases that I can think of and it seems to work well.

yajra:
Released on https://github.com/yajra/laravel-datatables/releases/tag/v12.1.1 🚀 Thanks!

PR #3228, opened 2025-04-22
feat: Public visibility for `resolveRelationColumn` method

Seb33300:
Alternative to https://github.com/yajra/laravel-datatables/pull/3229

Not sure which implementation is better, I will let you choose what you prefer.

I would like to suggest to make `resolveRelationColumn()` a public method so we can reuse it in our custom filters:

```php

$datatable
    ->filter(function ($query) use ($datatable) {
        // resolveRelationColumn automatically create joins if needed
        $column = $datatable->resolveRelationColumn('foo.bar');
        $query->where($column, 'xxx');
    })
    ->filterColumn('foo.bar', function ($query, $keyword) use ($datatable) {
        // resolveRelationColumn automatically create joins if needed
        $column = $datatable->resolveRelationColumn('foo.bar');
        $query->where($column, 'xxx');
    })
;
```

### DIFF
--- a/src/EloquentDataTable.php
+++ b/src/EloquentDataTable.php
@@ -155,12 +155,11 @@ class EloquentDataTable extends QueryDataTable
     }
 
     /**
-     * Resolve the proper column name be used.
-     *
+     * {@inheritDoc}
      *
      * @throws \Yajra\DataTables\Exceptions\Exception
      */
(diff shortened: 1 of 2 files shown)


Seb33300:
Since https://github.com/yajra/laravel-datatables/pull/3229 is merged, this on is no longer needed.

PR #3229, opened 2025-04-22
feat: add relation resolver param to filter callbacks

Seb33300:
Alternative to https://github.com/yajra/laravel-datatables/pull/3228

Not sure which implementation is better, I will let you choose what you prefer.

I would like to suggest to add resolveRelationColumn() as new param to the filter callbacks:

```php

$datatable
    ->filter(function ($query, $scoutSearch, $resolver) {
        // $resolver automatically create joins if needed
        $query->where($resolver('foo.bar'), 'xxx');
    })
    ->filterColumn('foo.bar', function ($query, $keyword, $resolver) {
        // $resolver automatically create joins if needed
        $query->where($resolver('foo.bar'), 'xxx');
    })
;
```

### DIFF
--- a/src/EloquentDataTable.php
+++ b/src/EloquentDataTable.php
@@ -155,8 +155,7 @@ class EloquentDataTable extends QueryDataTable
     }
 
     /**
-     * Resolve the proper column name be used.
-     *
+     * {@inheritDoc}
      *
      * @throws \Yajra\DataTables\Exceptions\Exception
      */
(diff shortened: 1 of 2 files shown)


yajra (approved):

yajra:
I like the resolver idea, thanks!

yajra:
Released on https://github.com/yajra/laravel-datatables/releases/tag/v12.1.0 🚀 Thanks!

PR #3232, opened 2025-05-05
feat: add relation resolver param to order callback

Seb33300:
Same as https://github.com/yajra/laravel-datatables/pull/3229 but for `orderColumn()`

In https://github.com/yajra/laravel-datatables/pull/3216, I added the `$column` parameter to pass the current column alias to the order callback function, but this is not really helpful when we want to work on other columns as reported by @Arkhas.

So I decided to replace this param by the resolver in the same way as `filterColumn()`

```php

$datatable
    ->orderColumn('foo.bar', function ($query, $order, $resolver) {
        // $resolver automatically create joins if needed
        $query->orderBy($resolver('foo.bar'), $order);
        // or
        $query->orderBy(DB::raw("{$resolver('foo.amount')} * {$resolver('foo.quantity')}"), $order);
    })
;
```

### DIFF
--- a/src/QueryDataTable.php
+++ b/src/QueryDataTable.php
@@ -691,11 +691,10 @@ class QueryDataTable extends DataTableAbstract
             })
             ->reject(fn ($orderable) => $this->isBlacklisted($orderable['name']) && ! $this->hasOrderColumn($orderable['name']))
             ->each(function ($orderable) {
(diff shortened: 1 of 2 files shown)


yajra:
Release on https://github.com/yajra/laravel-datatables/releases/tag/v12.2.0 🚀 Thanks!

PR #3234, opened 2025-05-07
feat: add option to enable alias on relation tables

Seb33300:
In https://github.com/yajra/laravel-datatables/pull/3216, I implemented automatic unique aliases on relation tables to prevent duplicated table name issues.

That bugfix unfortunately introduce a BC break that we cannot prevent.

As [suggested](https://github.com/yajra/laravel-datatables/pull/3216#issuecomment-2853852152) by @Arkhas, that makes that feature optional and disabled by default to keep backward compatibility.

```php
$datatable = new EloquentDataTable($this->query());
$datatable->enableEagerJoinAliases();
```

### DIFF
--- a/src/EloquentDataTable.php
+++ b/src/EloquentDataTable.php
@@ -18,6 +18,12 @@ use Yajra\DataTables\Exceptions\Exception;
 class EloquentDataTable extends QueryDataTable
 {
     /**
+     * Flag to enable the generation of unique table aliases on eagerly loaded join columns.
+     * You may want to enable it if you encounter a "Not unique table/alias" error when performing a search or applying ordering.
+     */
+    protected bool $enableEagerJoinAliases = false;
+
+    /**
      * EloquentEngine constructor.
      */
     public function __construct(Model|EloquentBuilder $model)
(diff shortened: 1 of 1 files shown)


Arkhas:
>Sounds good, maybe add the ability to set a default value in the config ? that's would be nice to have, but we can work with this if you don't think that would be a good idea.
>
>Thank you very much !

Seb33300:
@yajra Since it's not possible to implement alias on join table in a backward compatible way and because this is causing multiple bug reports and complication for developers who want to upgrade (see https://github.com/yajra/laravel-datatables/pull/3216#issuecomment-2853852152 or https://github.com/yajra/laravel-datatables/issues/3237#issuecomment-2884161372), I finally updated that PR to disable aliases on join tables by default.

The developer will have to opt in for that feature if he is facing a `Not unique table/alias` error:
```php
$datatable = new EloquentDataTable($this->query());
$datatable->enableEagerJoinAliases();
```

src/EloquentDataTable.php
```diff
@@ -17,6 +17,12 @@
  */
 class EloquentDataTable extends QueryDataTable
 {
+    /**
+     * Flag to enable the generation of unique table aliases on eagerly loaded join columns.
+     * You may want to enable it if you encounter a "Not unique table/alias" error when performing a search or applying ordering.
+     */
+    protected bool $enableEagerJoinAliases = false;
```
Seb33300, line 24: I was also thinking to name it `$uniqueEagerJoin`.

yajra (approved):

yajra:
Confirmed working as expected 💯, Thanks!

yajra:
Released on https://github.com/yajra/laravel-datatables/releases/tag/v12.3.0 🚀 Thanks!

Issue #3235, opened 2025-05-07
Column not found for version 12.1.1 and 12.1.2


alexbog8:
Hi,

When I use version 12.1.1 or version 12.1.2 I get column not found , unknown column in order clause. Version 12.1.0 and below works just fine.
What can be the problem? thanks.

alexbog8:
So I have something like this on the querybuilder

DB::raw(' ( CASE
        WHEN `end` >= CURDATE() THEN (DATEDIFF(`end`, CURDATE()) )
        WHEN `end` < CURDATE() THEN (DATEDIFF(`end`, CURDATE()) - DATEDIFF(`end`, CURDATE()) - DATEDIFF(`end`, CURDATE()) + '.$diff_in_days.' )
        END ) as diff_order')

then javascript
{ data: 'date', name: 'diff_order', 'searchable': false} and

order: [[6, "asc"]], 

In version 12.1.0 this works, but from version 12.1.1 no, with the error column not found workshops.diff_order

The quick fix is to order in the querybuilder and remove in javascript order: [[6, "asc"]],

yajra:
Always alias change was introduced in https://github.com/yajra/laravel-datatables/pull/3227 but I think we overlooked the use case when using raw SQL. 

Tagging for review and @Seb33300.

Thanks!

yajra:
For a workaround, maybe try using [`orderColumn`](https://yajrabox.com/docs/laravel-datatables/12.0)

Seb33300:
It looks like the alias is not properly detected.
I will try to reproduce tomorrow.

alexbog8:
This is the query builder

`$data = Workshop::select(['*',
            'workshops.id as id',
            DB::raw(' ( CASE
        WHEN `end` >= CURDATE() THEN (DATEDIFF(`start`, CURDATE()) )
        WHEN `end` < CURDATE() THEN (DATEDIFF(`start`, CURDATE()) - DATEDIFF(`start`, CURDATE()) - DATEDIFF(`start`, CURDATE()) + '.$diff_in_days.' )
        END ) as diff_order, ( CASE 
             WHEN `end` >= CURDATE() THEN (DATEDIFF(`end`, CURDATE()) )
             WHEN `end` < CURDATE() THEN (DATEDIFF(`start`, CURDATE()) - DATEDIFF(`start`, CURDATE()) - DATEDIFF(`start`, CURDATE()) + '.$diff_in_days.' ) 
[...]

PR #3236, opened 2025-05-08
fix: improve column alias detection

Seb33300:
Fixes https://github.com/yajra/laravel-datatables/issues/3235

This PR improves detection of column aliases.

Before, it could detect only if the alias has exactly 1 space between `as` and alias name.
Now it supports (I hope) all cases:
```php
DB::raw('table.column as   alias'); // multiple spaces
DB::raw('table.column as `alias`'); // escaped (wrapped) aliases
DB::raw('table.column as alias, table.column2 as alias2'); // multiple aliases in 1 definition
// multiple lines
DB::raw('table.column as  
alias');
```


yajra (approved):

yajra:
Release on https://github.com/yajra/laravel-datatables/releases/tag/v12.2.0 🚀 Thanks!

Issue #3237, opened 2025-05-08
Sorting results in error when prefix of column is not defined


tkcsdemos:
From version 12.1.1 and after we have encountered an error on sorting on columns without a table prefix. Although the sorting columns are unique between the joined tables the sorting results in error because the prefix is enforced wrongly to the primary table.

Example eloquent query:

```php
$query = Member::select(['members.*',
            'qprofiles.id as qprofiles_id',
            'qprofiles.active as qprofiles_active',
            'qprofiles.date',
            'qprofiles.updated_at',
            'qprofiles.vision',
            'qprofiles.talent',
            'qprofiles.competence',
            'qprofiles.performance',
            'qprofiles.attrition',
            'qprofiles.operating',
            'qprofiles.profile_role',
            'qprofiles.score',
            'qprofiles.smart_questions',
            'qprofiles.time_in_role',
            'qprofiles.reward',
            'qprofiles.development_scan',
            'qprofiles.organization_design_issue',
            'qprofiles.campaign_id',
        ])
            ->with(['department', 'fields'])
            ->join('departments', 'members.department_id', '=', 'departments.id')
            ->leftjoin('qprofiles', function ($join) {
                $join->on('members.id', '=', 'qprofiles.member_id');
                $join->on('qprofiles.active', '=', DB::raw('1'));
            })
            ->where('departments.organization_id', Auth::user()->organization_id)
            ->where('members.active', 1);

return DataTables::of($query )
            ->addIndexColumn()
            ->setRowId('id')
...
```

Seb33300:
Hi @tkcsdemos 

Can you please share how you defined columns and the error message?

tkcsdemos:
> Hi [@tkcsdemos](https://github.com/tkcsdemos)
> 
> Can you please share how you defined columns and the error message?

Of course. I defined columns on the backend as shown below:

```php
return DataTables::of($query)
  ->addIndexColumn()
  ->setRowId('id')
  ...,
  ->editColumn('date', function ($row) {
      return $row->date ? date('d-m-Y', strtotime($row->date)) : '';
  })
  ->smart()
  ->make();
```

and on the frontend:

```js
let tableMembers = $('#membersTbl').DataTable({
    ajax: {
        url: "{{ route('my_route') }}",
        type: 'POST',
        headers: {'X-CSRF-TOKEN': "{{ csrf_token() }}"},
    },
    serverSide: true,
    pageLength: 100,
    responsive: false,
[...]

PR #3238, opened 2025-05-09
fix: improve prefix detection

Seb33300:
Fixes https://github.com/yajra/laravel-datatables/issues/3237

In the previous implementation, I was looking only for prefixed columns.
Unfortunately, this was not working as expected when custom select without alias were manually added.

I have refactored to try to retrieve the table prefix by searching in all selected columns even if they do not have an alias.

### DIFF
--- a/src/QueryDataTable.php
+++ b/src/QueryDataTable.php
@@ -475,20 +475,26 @@ class QueryDataTable extends DataTableAbstract
             return $column;
         }
 
-        $q = $this->getBaseQueryBuilder($query);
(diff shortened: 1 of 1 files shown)


yajra (approved):
It would be nice if we can have unit tests for this cases. Will also try to implement when I get the chance.

Good to go as tested by @tkcsdemos

yajra:
Released on https://github.com/yajra/laravel-datatables/releases/tag/v12.2.1 🚀

Thanks @Seb33300 and @tkcsdemos

PR #3243, opened 2025-06-09
fix: support for array notation

Seb33300:
See https://datatables.net/reference/option/columns.data#string

It's possible to specify the separator of array values inside the brackets: `name[, ]`

Before this patch, only `name[]` was working.


yajra:
Released on https://github.com/yajra/laravel-datatables/releases/tag/v12.3.1 🚀 Thanks!

Issue #3249, opened 2025-09-10
Bug when filtering words with accents


pabloroxo:
Hi, Arjay!

I'm facing a bug when fitering by term.

For example: I list users and there is a user called "Tatiane Simões". Note the "õ". If I search for "simões", it will show in the list. But sometimes we don't type the accents. We just put "simoes" in the field and expect that "Tatiane Simões" will be listed. But it's not.

Is there any way we could ignore these accents when filtering? Maybe adding some boolean option, like in other optios in "/vendor/yajra/laravel-datatables-oracle/src/config/datatables.php". If yes, here's a list of letters with accents in portuguese brazilian which we'd love to be replaced:

Ã/ã/Á/á/À/à/Â/â = a
É/é/Ê/ê = e
Í/í = i
Ó/ó/Ô/ô/Õ/õ = o
Ú/ú = u
Ç/ç = c

We don't use other combinations of letter and accents.

Thank you!

pabloroxo:
So, my suggestion is that you implement replacing characters, for example from "õ" to "o" and "Õ" to "o" (considering everything is transformed to lowercase) BEFORE searching for the substring in "search[value]", passed through the request:

<img width="215" height="25" alt="Image" src="https://github.com/user-attachments/assets/a5a45096-b7be-46ee-a0e6-36688d9de92c" />

Accents are used in many many languages, even in english, with some words like "fiancé", "resumé" and "façade".

Or if you don't consider it a feature to be developed, maybe give me instructions for where I can implement it.

Thank you!

pabloroxo:
Well, I solved this by editing your code (`/vendor/yajra/laravel-datatables-oracle/src/CollectionDataTable.php`):

<img width="1021" height="1141" alt="Image" src="https://github.com/user-attachments/assets/0022f596-13d3-4551-a6ea-b927aaa8046a" />

I know it's an HORRIBLE solution, even using a deprecated function in PHP 8.2 (although I'm using PHP 8.0), but... it works!

<img width="1665" height="519" alt="Image" src="https://github.com/user-attachments/assets/e1f3e451-17a0-47bf-b24a-012ab3b7cd2f" />

I'll leave this a suggestion for a next release!

Thanks!

pabloroxo:
Nope. Nothing related to collation. As I said on [this comment](https://github.com/yajra/laravel-datatables/issues/3249#issuecomment-3376859528):

> So, it's not a matter of collation.

[...]

PR #3260, opened 2025-10-15
Implement accent-insensitive search for Portuguese characters

slowestwind:
## Description

This PR implements accent-insensitive search functionality to address the issue where users cannot find results when searching without accents (e.g., searching 'simoes' to find 'Simões').

## Problem

Users in Portuguese-speaking countries often don't type accents when searching, but expect to find results with accented characters. This is especially common in Brazilian Portuguese where users might search for:
- 'simoes' instead of 'Simões'
- 'joao' instead of 'João'
- 'sao paulo' instead of 'São Paulo'

## Solution

Added a new configuration option `ignore_accents` that when enabled, normalizes both search terms and data to their base characters before comparison.

### Changes Made

1. **Configuration**: Added `'ignore_accents' => false` to the search config in `datatables.php`
2. **Helper Method**: Created `Helper::normalizeAccents()` to convert accented characters to base characters
3. **Config Method**: Added `Config::isIgnoreAccents()` to check if the feature is enabled
4. **Query DataTable**: Updated to use database functions for accent normalization in SQL queries
5. **Collection DataTable**: Updated to normalize both search terms and data values
6. **Tests**: Added comprehensive unit tests for the normalization function

### Supported Characters (Portuguese Brazilian)

- Ã/ã/Á/á/À/à/Â/â → a
- É/é/Ê/ê → e  
- Í/í → i
- Ó/ó/Ô/ô/Õ/õ → o
- Ú/ú → u
- Ç/ç → c

## Usage

Enable the feature in your config:

```php
'search' => [
    'ignore_accents' => true,
    // other settings...
],
```

When enabled:
- Searching for 'simoes' will find 'Simões'
- Searching for 'joao' will find 'João'  
- Searching for 'sao' will find 'São'

## Testing

- Added unit tests covering all accent mappings
- Tests verify both individual character replacement and full text normalization
- Backwards compatible - feature is disabled by default

## Database Compatibility

The feature uses appropriate database functions:
- **MySQL**: `REPLACE()` function cascade for each accent mapping
- **PostgreSQL**: `UNACCENT()` function if available, fallback to `REPLACE()`
- **SQLite**: `REPLACE()` function cascade
- **SQL Server**: `REPLACE()` function cascade

## Breaking Changes

None - this is a new optional feature that is disabled by default.

Fixes yajra/laravel-datatables#3249

### DIFF
--- a/src/QueryDataTable.php
+++ b/src/QueryDataTable.php
@@ -564,6 +564,11 @@ class QueryDataTable extends DataTableAbstract
      */
     protected function compileQuerySearch($query, string $column, string $keyword, string $boolean = 'or'): void
     {
+        // Validate inputs to prevent any potential issues
+        if (empty($column) || empty($keyword)) {
+            return;
+        }
+
         $wrappedColumn = $this->wrap($this->addTablePrefix($query, $column));
         $castedColumn = $this->castColumn($wrappedColumn);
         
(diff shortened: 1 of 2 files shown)


Seb33300:
Hi @slowestwind, which database are you using?
To support accent-insensitive search, the proper way is to configure your database with an appropriate collation.

With MySQL, this can be done using collations such as `utf8mb4_unicode_ci`.
With PostgreSQL, this is usually achieved by using a suitable collation together with the `unaccent` extension in your queries.

PR #3299, opened 2026-08-12
perf: group global search columns of the same relation into one sub-query

### LINKED ISSUE #2519: Global Search performance

Hi all, I've been using this package for a few months now and it's been working great for my requirements !

This weekend though, during some patches to boost the performance of my app, I've noticed that a few queries were taking a bit longer than I expected, and I think I managed to find what was causing it.

I'm not sure if this is intentional or not, but when looking through the source code of the package, I saw that on the `globalSearch()` method, we don't call the `resolveRelationColumn()` like we do on the column search and ordering.

The result of this is that it adds a subquery for each column of a relationship that is available on the global search instead of performing the join. On a table with just a few thousands of records, the SQL ran 3x slower than it did with the join.

### Code snippet of problem

Say that you have an instance of the DataTables for a User model that has a relation to an "Employee" model, which stores all the basic information, such as first and last names, address, etc.

Now you want to allow the user to perform the global search with any of the parameters of the `Employee` relationship, this will add a subquery for each attribute, for example:

```sql
where (
    exists (select * from `employees`
        where `users`.`employee_id` = `employees`.`id`
            and LOWER(`employees`.`first_name`) LIKE '%ian%'
    )
    or exists (select * from `employees`
        where `users`.`employee_id` = `employees`.`id`
            and LOWER(`employees`.`last_name`) LIKE '%ian%'
    )
)
```

If we add the line
```php
$column  = $this->resolveRelationColumn($column);
```

Right before the following one:
https://github.com/yajra/laravel-datatables/blob/7e45885588225b31a9e7bdf4b04733c44432531c/src/QueryDataTable.php#L751

We get a much nicer and faster query (with a single sub-query per relationship):

```sql
where (
    exists (select * from `employees`
        where `users`.`employee_id` = `employees`.`id`
            and (LOWER(`employees`.`first_name`) LIKE '%ian%' or LOWER(`employees`.`last_name`) LIKE '%ian%')
    )
)
```

So my question here is: is this intentional and we can't do this for the globalSearch or is it just something that went unnoticed?

I'm glad to open a PR with the patch if needed.

### System details

- Operating System: MacOS 10.15.7
- PHP Version: 7.4
- Laravel Version: 7.28.1
- Laravel-Datatables Version: 9.11.0

OzanKurt:
Closes #2519

## Problem

Global search compiles one query per searchable column. For an eager loaded relation that means one `exists` sub-query per column, all hitting the same table:

```sql
where (
    exists (select * from "users" where "posts"."user_id" = "users"."id" and lower("users"."name") like ?)
    or exists (select * from "users" where "posts"."user_id" = "users"."id" and lower("users"."email") like ?)
)
```

Column search and ordering already resolve relation columns once, so global search was the odd one out, and the cost grows with every searchable column of the relation.

## Solution

Searchable columns are now grouped per relation before being compiled, so each relation produces a single sub-query:

```sql
where exists (
    select * from "users"
    where "posts"."user_id" = "users"."id"
    and (lower("users"."name") like ? or lower("users"."email") like ?)
)
```

The column conditions are wrapped in a nested `where` so the `or` cannot leak out of the relation constraint. That wrapper is only added when a relation has more than one searchable column, keeping the generated SQL identical to before for every single column case.

The original suggestion in the issue was to call `resolveRelationColumn()` in `globalSearch()`. That turns the sub-query into a join, which also multiplies rows for `hasMany` / `belongsToMany` relations and changes `recordsFiltered`. Grouping keeps the current semantics and only removes the duplicated sub-queries, which is what was agreed on in the discussion.

## Changes

- `QueryDataTable::globalSearch()` now delegates to `compileGlobalSearch()`, which loops the searchable columns through `compileGlobalSearchColumn()`. Behaviour is unchanged for the query engine.
- `EloquentDataTable::compileGlobalSearch()` groups the columns by relation, preserving the order of first appearance, and compiles each group into one `whereHas` / `whereHasMorph`.
- `EloquentDataTable::compileQuerySearch()` was split into `resolveSearchableRelation()` (which relation, if any, a column belongs to) and `compileRelationSearch()` (compile one relation with n columns), so single column search and grouped global search share one code path.

## Notes

- Not affected: column search, ordering, `filterColumn()` callbacks, columns of relations that are not eager loaded, and morph relations keep their existing per relation handling.
- `tests/Unit/GlobalSearchRelationTest.php` covers the grouping, one sub-query per relation, the untouched non eager loaded case, the `and (... or ...)` precedence, and that the result set is unchanged. Three of these tests fail on the current implementation.
- `composer stan` passes and the existing relation tests (`BelongsTo`, `HasMany`, `HasOne`, `HasOneThrough`, `BelongsToMany`, `MorphTo`, `DeepRelation`, `ArrayNotationRelation`) pass unchanged.

### DIFF
--- a/src/EloquentDataTable.php
+++ b/src/EloquentDataTable.php
@@ -13,6 +13,7 @@ use Illuminate\Database\Eloquent\Relations\MorphTo;
 use Illuminate\Database\Eloquent\Relations\Relation;
 use Illuminate\Database\Query\Builder as BaseQueryBuilder;
 use Illuminate\Database\Query\JoinClause;
+use Illuminate\Support\Collection;
 use Yajra\DataTables\Exceptions\Exception;
 
 /**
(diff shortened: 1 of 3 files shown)


yajra:
Please resolve the conflicts, thanks!

OzanKurt:
Rebased on v13.2.0 and resolved, the conflict was only in the `use` block of `EloquentDataTable.php` against #3303. Tests and static analysis are green.

OzanKurt:
Heads up on an interaction with #3300: both PRs rewrite `EloquentDataTable::compileQuerySearch()`, and they were branched off master independently.

#3300 adds `resolveRelationName()` calls inside `compileQuerySearch()`, while this one replaces that method with `resolveSearchableRelation()`, which does not call it. Merging both conflicts in that method. Resolving the conflict by keeping the version from this PR compiles and looks right, but the snake case resolution then only applies to ordering and is lost for search. `SnakeCaseRelationTest` fails in that case, so it will not slip through, but it is easier to just resolve it correctly.

The correct resolution is to call `resolveRelationName()` at both return points of `resolveSearchableRelation()`:

```php
$relation = $this->resolveRelationName(array_shift($parts));
// ...
$relation = $this->resolveRelationName(implode('.', $parts));
```

I merged both branches locally with that resolution and the suite is green, 182 tests, PHPStan clean.

Whichever of the two you merge first, I will push the follow up to the other one right away, so there is no need to resolve it by hand.

PR #3300, opened 2026-08-13
fix: resolve eager loaded relations written in snake case

### LINKED ISSUE #2324: global search fails when column name with relation has underscore
Global search failing because EagerLoads collection contains relation names in CamelCase, whereas the founction is passing in the actual table name from the column name. e.g.

column name =  child_table.name

Name in EagerLoad collection = childTable

compileQuerySearch in EloquentDataTable.php fails because it needs to convert relation name to camcel Case before calling isNotEagerLoaded:

suggest adding Str::camel call as follows:

```php
protected function compileQuerySearch($query, $columnName, $keyword, $boolean = 'or')
    {
        $parts    = explode('.', $columnName);
        $column   = array_pop($parts);
        $relation = Str::camel( implode('.', $parts) );    // Added convert to camelCase

        if ($this->isNotEagerLoaded( $relation )) {
            return parent::compileQuerySearch($query, $columnName, $keyword, $boolean);
        }

        $query->{$boolean . 'WhereHas'}($relation, function (Builder $query) use ($column, $keyword) {
            parent::compileQuerySearch($query, $column, $keyword, '');
        });
    }
```

Same issue could be fixed in resolveRelationColumn: 

```php
protected function resolveRelationColumn($column)
    {
        $parts      = explode('.', $column);
        $columnName = array_pop($parts);
        $relation   = Str::camel( implode('.', $parts) );   // Added convert to camelCase

        if ($this->isNotEagerLoaded($relation)) {
            return $column;
        }

        return $this->joinEagerLoadedColumn($relation, $columnName);
    }
```

OzanKurt:
Closes #2324

## Problem

Eager loaded relations are matched against the column name as given, but relations are defined in camel case while column names are usually written in snake case. A column named `child_table.name` is looked up as the relation `child_table`, which is never in `getEagerLoads()`, so the column falls through to the plain query search and produces an unknown column or missing table error.

The workaround is to write the name twice on the JS side:

```js
{data: 'child_table.name', name: 'childTable.name'}
```

## Solution

When the relation taken from the column name is not eager loaded, its camel case form is used if that one is. As suggested in the issue discussion, this makes both spellings work:

```js
{data: 'child_table.name', name: 'childTable.name'} // still works
{data: 'child_table.name'}                          // now works too
```

The fallback only applies when the literal name is not an eager loaded relation, so a relation genuinely defined as `child_table()` keeps resolving to itself and nothing that works today changes.

## Changes

- `EloquentDataTable::resolveRelationName()` resolves a relation name against the eager loads, falling back to `Str::camel()`.
- Applied in `compileQuerySearch()` (global and column search, including nested relations) and `resolveRelationColumn()` (ordering and joins), so search and ordering behave the same way.
- `tests/Integration/SnakeCaseRelationTest.php` covers global search, column search, and ordering on a snake case column name, plus a regression test that a relation resolving under its literal name is untouched. Three of these fail on the current implementation.

## Notes

- Nested relations are handled as well, since `Str::camel('child_table.sub_table')` resolves to `childTable.subTable`.
- A `postUser` relation was added to the `Post` test model to have a relation whose camel case name differs from the column name.
- `composer stan` passes and the existing test suite is unchanged.

### DIFF
--- a/src/EloquentDataTable.php
+++ b/src/EloquentDataTable.php
@@ -13,6 +13,7 @@ use Illuminate\Database\Eloquent\Relations\MorphTo;
 use Illuminate\Database\Eloquent\Relations\Relation;
 use Illuminate\Database\Query\Builder as BaseQueryBuilder;
 use Illuminate\Database\Query\JoinClause;
+use Illuminate\Support\Str;
 use Yajra\DataTables\Exceptions\Exception;
 
 /**
@@ -92,7 +93,7 @@ class EloquentDataTable extends QueryDataTable
             }
 
             $parts = explode('.', $column);
-            $firstRelation = array_shift($parts);
+            $firstRelation = $this->resolveRelationName(array_shift($parts));
             $column = implode('.', $parts);
 
             if ($this->isMorphRelation($firstRelation)) {
@@ -114,7 +115,7 @@ class EloquentDataTable extends QueryDataTable
 
         $parts = explode('.', $column);
         $newColumn = array_pop($parts);
-        $relation = implode('.', $parts);
+        $relation = $this->resolveRelationName(implode('.', $parts));
 
         if (! $nested && $this->isNotEagerLoaded($relation)) {
             parent::compileQuerySearch($query, $column, $keyword, $boolean);
@@ -135,6 +136,24 @@ class EloquentDataTable extends QueryDataTable
                 parent::compileQuerySearch($query, $newColumn, $keyword, '');
             });
         }
+    }
+
+    /**
+     * Resolve the name of an eager loaded relation.
+     *
+     * Column names are usually written in snake case, e.g. "child_table.name",
+     * while the relation itself is defined in camel case. The camel case
+     * relation is therefore used when it is the eager loaded one.
+     */
+    protected function resolveRelationName(string $relation): string
+    {
+        if (! $relation || array_key_exists($relation, $this->query->getEagerLoads())) {
+            return $relation;
+        }
+
+        $camel = Str::camel($relation);
+
+        return array_key_exists($camel, $this->query->getEagerLoads()) ? $camel : $relation;
     }
 
     /**
@@ -190,7 +209,9 @@ class EloquentDataTable extends QueryDataTable
     {
         $parts = explode('.', $column);
         $columnName = array_pop($parts);
-        $relation = preg_replace('/\[.*?\]/', '', implode('.', $parts));
+        $relation = $this->resolveRelationName(
+            (string) preg_replace('/\[.*?\]/', '', implode('.', $parts))
+        );
 
         if ($this->isNotEagerLoaded($relation)) {
             return parent::resolveRelationColumn($column);
--- a/src/EloquentDataTable.php
+++ b/src/EloquentDataTable.php
@@ -167,9 +167,52 @@ class EloquentDataTable extends QueryDataTable
             return $relation;
         }
 
-        $camel = Str::camel($relation);
+        $resolved = null;
+        $resolvedScore = -1;
 
-        return array_key_exists($camel, $this->query->getEagerLoads()) ? $camel : $relation;
+        foreach (array_keys($this->query->getEagerLoads()) as $eagerRelation) {
+            $score = $this->relationNameMatchScore($relation, (string) $eagerRelation);
+
+            if ($score !== null && $score > $resolvedScore) {
+                $resolved = (string) $eagerRelation;
+                $resolvedScore = $score;
+            }
+        }
+
+        return $resolved ?? $relation;
+    }
+
+    /**
+     * Score how well a relation matches an eager loaded one, segment by segment.
+     *
+     * Null means the two cannot be the same relation, otherwise the score is the
+     * number of segments that matched literally, so that an eager load spelled
+     * exactly like the column wins over one that only matches in camel case.
+     */
+    protected function relationNameMatchScore(string $relation, string $eagerRelation): ?int
+    {
+        $parts = explode('.', $relation);
+        $eagerParts = explode('.', $eagerRelation);
+
+        if (count($parts) !== count($eagerParts)) {
+            return null;
+        }
+
+        $score = 0;
+
+        foreach ($parts as $index => $part) {
+            if ($part === $eagerParts[$index]) {
+                $score++;
+
+                continue;
+            }
+
+            if (Str::camel($part) !== $eagerParts[$index]) {
+                return null;
+            }
+        }
+
+        return $score;
     }
 
     /**
@@ -177,13 +220,13 @@ class EloquentDataTable extends QueryDataTable
      */
     protected function resolveRelationNameOf(Model $model, string $relation): string
     {
-        if (method_exists($model, $relation)) {
+        if ($model->isRelation($relation)) {
             return $relation;
         }
 
         $camel = Str::camel($relation);
 
-        return method_exists($model, $camel) ? $camel : $relation;
+        return $model->isRelation($camel) ? $camel : $relation;
     }
 
     /**
--- a/src/EloquentDataTable.php
+++ b/src/EloquentDataTable.php
@@ -4,6 +4,7 @@ namespace Yajra\DataTables;
 
 use Illuminate\Contracts\Database\Eloquent\Builder as EloquentBuilder;
 use Illuminate\Contracts\Database\Query\Builder as QueryBuilder;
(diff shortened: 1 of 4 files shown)


yajra:
This is a great fix @OzanKurt, love it! I personally encounter similar concern of forgetting camel and snake case 😅

Please resolve the conflicts when you get the chance.

Thanks!

OzanKurt:
Thanks! Rebased on v13.2.0 and resolved, the conflict was only in the `use` block of `EloquentDataTable.php` against #3303. Tests and static analysis are green.

OzanKurt:
Heads up on an interaction with #3299: both PRs rewrite `EloquentDataTable::compileQuerySearch()`, and they were branched off master independently.

This PR adds `resolveRelationName()` calls inside `compileQuerySearch()`, while #3299 replaces that method with `resolveSearchableRelation()`, which does not call it. Merging both conflicts in that method. Resolving the conflict by keeping the version from #3299 compiles and looks right, but the snake case resolution then only applies to ordering and is lost for search. `SnakeCaseRelationTest` fails in that case, so it will not slip through, but it is easier to just resolve it correctly.

The correct resolution is to call `resolveRelationName()` at both return points of `resolveSearchableRelation()`:

```php
$relation = $this->resolveRelationName(array_shift($parts));
// ...
$relation = $this->resolveRelationName(implode('.', $parts));
```

I merged both branches locally with that resolution and the suite is green, 182 tests, PHPStan clean.

Whichever of the two you merge first, I will push the follow up to the other one right away, so there is no need to resolve it by hand.

yajra:
Thanks @OzanKurt — great fix, and verified working at depth 1 (search + order + column search) against the head. One issue before merge:

**Nested snake_case relations are still broken for search.** `compileQuerySearch()` only resolves the first segment (`array_shift`) against the root builder's eager loads; deeper segments recurse as bare names that don't exist in `getEagerLoads()` (Laravel stores dotted paths like `user.nestedFilteredHeart`, not bare keys), so `whereHas()` receives the raw snake name.

Repro on this branch: with `Post::with('user.nestedFilteredHeart')` and column `user.nested_filtered_heart.size`, global search returns:

```php
Call to undefined method Yajra\DataTables\Tests\Models\User::nested_filtered_heart()
```

PS: code reviewed via Codex and Ox Alpha

OzanKurt:
Good catch, thanks. Reproduced exactly, including the `User::nested_filtered_heart()` fatal.

One refinement: it is specifically the `whereHas` recursion, so only global search was affected. Nested column search and nested ordering already worked, because they go through `resolveRelationColumn()`, which camel cases the whole dotted path, and `Str::camel('user.nested_filtered_heart')` gives `user.nestedFilteredHeart`, which does match the eager load key. I added tests for all three and only the two global search ones were red.

Fixed by resolving a nested segment against the model that declares it instead of against the root builder's eager loads. The root level still resolves against the eager loads, so depth 1 behaviour is unchanged.

It also fixes a case that never worked: a nested relation that is not eager loaded at the deeper level, e.g. `Post::with('user')` with `user.nested_filtered_heart.size`. That is consistent with `$nested = true` already skipping the eager load check.

Added 4 tests: nested global search, nested column search, nested ordering, and the not eager loaded nested case. Two of them fail on the previous head.

Unrelated pre-existing limitation I noticed while in there, not touched here: `isMorphRelation()` always checks the root model, so a nested morph relation is not detected as one. Happy to send that separately if you want it.

OzanKurt:
Correcting myself on the morph note above: I had the direction backwards.

A nested morph relation not being detected is harmless. Laravel's `has()` already routes a `MorphTo` to `hasMorph($relation, '*')`, so `whereHas()` produces the same per type existence queries with the `_type` constraints.

The actual problem is the opposite one. Because `isMorphRelation()` resolves against the root model, a plain nested relation that happens to share its name with a morph relation of the root model is sent to `whereHasMorph()` and fatals:

```
Call to undefined method Illuminate\Database\Eloquent\Relations\HasOne::getMorphType()
```

With the test models, `User::user()` is a `morphTo` while `Post::user()` is a `belongsTo`, so searching `posts.user.name` on `User::with('posts.user')` hits it.

Sent as #3306, stacked on this branch since it reuses the nested context added here. Both directions are covered by tests there.

yajra:
## 2 issues:

### 1. SQL Error on mixed literal and camelCase path

```php
// Post.php — this relation is literally snake_case
public function post_user(): BelongsTo
{
    return $this->belongsTo(User::class, 'user_id');
}

// User.php — this nested relation is camelCase
public function nestedFilteredHeart(): HasOne
{
    return $this->hasOne(Heart::class);
}

$query = Post::with('post_user.nestedFilteredHeart')
    ->select('posts.*');

return (new EloquentDataTable($query))->toJson();

// DataTables request
columns[0][data]=post_user.nested_filtered_heart.size
columns[0][name]=post_user.nested_filtered_heart.size
order[0][column]=0&order[0][dir]=desc
```

**Exception**

```bash
Exception Message:

SQLSTATE[HY000]: General error: 1 no such column: post_user.nested_filtered_heart.size (Connection: sqlite, Database: /Users/yajra/code/packages/dataTables/pr-3300-reproduction/database/database.sqlite, SQL: select "posts".* from "posts" order by "post_user"."nested_filtered_heart"."size" desc limit 10 offset 0)
```

### 2. Dynamic relation resolver

```php
// AppServiceProvider.php
User::resolveRelationUsing(
    'dynamicHeartRelation',
    fn (User $user): HasOne => $user->hasOne(Heart::class),
);

$query = Post::with('post_user.dynamicHeartRelation')
    ->select('posts.*');

return (new EloquentDataTable($query))->toJson();

// DataTables request
columns[0][data]=post_user.dynamic_heart_relation.size
columns[0][name]=post_user.dynamic_heart_relation.size
search[value]=heart-2
```

**Exception**

```php
Exception Message:

Call to undefined method App\Models\User::dynamic_heart_relation()
```

## Proposed changes (EloquentDataTable.php)

Replace the PR implementation with segment-aware eager-load matching and Eloquent's own relation detection.

```php
use Illuminate\Database\Eloquent\Builder as BaseEloquentBuilder;
use Illuminate\Database\Eloquent\Model;
use Illuminate\Support\Str;

protected function resolveRelationName(string $relation, $query = null): string
{
    if (! $relation) {
        return $relation;
    }

    if ($query instanceof BaseEloquentBuilder) {
        return $this->resolveRelationNameOf($query->getModel(), $relation);
    }

    if (array_key_exists($relation, $this->query->getEagerLoads())) {
        return $relation;
    }

    $resolved = null;
    $resolvedScore = -1;

    foreach (array_keys($this->query->getEagerLoads()) as $eagerRelation) {
        $score = $this->relationNameMatchScore($relation, $eagerRelation);

        if ($score !== null && $score > $resolvedScore) {
            $resolved = $eagerRelation;
            $resolvedScore = $score;
        }
    }

    return $resolved ?? $relation;
}

protected function resolveRelationNameOf(Model $model, string $relation): string
{
    if ($model->isRelation($relation)) {
        return $relation;
    }

    $camel = Str::camel($relation);

    return $model->isRelation($camel) ? $camel : $relation;
}

private function relationNameMatchScore(
    string $relation,
    string $eagerRelation,
): ?int {
    $parts = explode('.', $relation);
    $eagerParts = explode('.', $eagerRelation);

    if (count($parts) !== count($eagerParts)) {
        return null;
    }

    $score = 0;

    foreach ($parts as $index => $part) {
        if ($part === $eagerParts[$index]) {
            $score++;

            continue;
        }

        if (Str::camel($part) !== $eagerParts[$index]) {
            return null;
        }
    }

    return $score;
}
```

OzanKurt:
Both reproduced, thanks. Fixed by adopting your implementation.

Segment aware matching against the eager loads with the literal spelling scoring higher, plus `Model::isRelation()` so relations registered through `resolveRelationUsing()` are found. Tests added for both, the mixed `post_user.nestedFilteredHeart` path when ordering and the dynamic relation when searching. Both are red on the previous head.

Two notes.

I kept `relationNameMatchScore()` as `protected` rather than `private`, since every other helper in the class is protected and the engine gets extended a fair bit. Happy to make it private if you prefer.

The same `method_exists()` gap was in the morph check, so a relation registered through `resolveRelationUsing()` would never be detected as a morph one. That is fixed in #3306, which is rebased on this branch.

Small thing on your second repro: after the fix it returns 6 rows rather than an error, because `%heart-2%` also matches `heart-20`. My test searches `heart-20` so the count is unambiguous.

yajra (approved):

PR #3301, opened 2026-08-13
feat: add a max_length config to cap the records per request

### LINKED ISSUE #1597: Default total records (length) to API
### Feature request
Hello, we are developing an api and we are certainly interested in using your component.

We have consulted the documentation but there is no element that allows us to define in the configuration the amount of elements to obtain by default.

The application is an API, it does not have direct interaction with a site or Javascript.

```php
'search' => [
        /*
         * Sets the amount of item to obtain by default.
         * It must be a positive integer.
         * 
         * 0: without limit
         */
        'length' => 20,
```
```php
datatables()->eloquent( $query )
            ->setDefaultLength(30)
            ->make()
```

Regards

### System details

- Amazon Linux AMI / Windows Server 2018
- PHP 7.0.21
- laravel/lumen-framework 5.5 
- illuminate/database 5.5

### LINKED ISSUE #2493: recordsFiltered Limit (Hard Limit in Config)

OzanKurt:
Closes #2493
Closes #1597

## Problem

There is no way to cap how many records a single request can pull. A client can send `length=-1`, or simply omit `start` and `length`, and get the whole table back. On a public endpoint that is an easy way to put load on the server, and it is the reason both issues were opened.

The suggested workaround was a middleware rewriting the request:

```php
if (request('length') > 100) {
    request()->merge(['length' => 100]);
}
```

That has to be repeated per project and still does not cover a request that sends no length at all.

## Solution

A `max_length` config option, `null` by default so nothing changes unless it is set:

```php
/*
 * Maximum number of records that can be requested per page.
 * A request asking for more, or for all the records using a length of -1,
 * is capped to this value. Set to null to allow any length.
 */
'max_length' => null,
```

When set, it caps every response:

| request | without max_length | with `max_length => 100` |
| --- | --- | --- |
| `length=25` | 25 records | 25 records |
| `length=5000` | 5000 records | 100 records |
| `length=-1` | all records | 100 records |
| no `start` / `length` | all records | 100 records |

The requested `start` is still honoured, so paging keeps working against the capped length.

## Opting out, for exports

You raised the export case, and it was a real regression: buttons forces `length => -1` in `getAjaxResponseData()` to collect every filtered record, so the cap would have truncated every export. There is now an escape hatch:

```php
app('datatables.request')->ignoreMaxLength(); // or $dataTable->ignoreMaxLength();
```

It is set on the request object rather than derived from the request input, so a client cannot send something to bypass the configured maximum. The matching one line change for buttons is yajra/laravel-datatables-buttons#214.

Two things were needed to make that work correctly:

- `Utilities\Request` is now aliased to `datatables.request`. The engines resolve the request from `datatables.request` while buttons resolves it from the class name, and those were two different instances, so the flag would never have reached the engine.
- The binding is `scoped` instead of `singleton`. `ignoreMaxLength()` is the first state ever held on the request object, and a shared instance would keep the cap disabled for every later request of a long running worker such as Octane.

## Changes

- `Request::maxLength()` reads the config value, returning 0 when unset or when ignored.
- `Request::length()` returns the maximum when the requested length is above it or is a "give me everything" value.
- `Request::isPaginationable()` returns true when a maximum is configured, so that a request with `length=-1` or no paging parameters at all is paginated instead of returning the full table.
- `Request::ignoreMaxLength()` and the fluent `DataTableAbstract::ignoreMaxLength()` opt a request out.
- `tests/Integration/MaxLengthTest.php` covers the unset default, a request with no parameters, `length=-1`, a length above and below the maximum, paging from a non zero start, the collection engine, both opt out entry points, that the request resolves to the same instance under both keys, and that the opt out does not survive into the next request.

## Notes

- Default is `null`, so the behaviour of existing applications is unchanged until they opt in.
- The cap lives in `Utilities\Request`, so every engine that pages through `request->length()` (eloquent, query, collection, resource, paginator) is covered by the same rule.
- `recordsTotal` and `recordsFiltered` still report the real counts, so the client side pager stays correct.
- `composer stan` passes and the existing test suite is unchanged.

### DIFF
--- a/src/DataTableAbstract.php
+++ b/src/DataTableAbstract.php
@@ -1044,6 +1044,21 @@ abstract class DataTableAbstract implements DataTable
         return $this;
     }
 
+    /**
+     * Ignore the maximum length configured via datatables.max_length.
+     *
+     * Use it when all the records are needed no matter the configured
+     * maximum, e.g. when exporting every filtered record.
+     *
+     * @return $this
+     */
+    public function ignoreMaxLength(): static
+    {
+        $this->request->ignoreMaxLength();
+
+        return $this;
+    }
+
     protected function validateMinLengthSearch(): void
     {
         if ($this->request->isSearchable()
--- a/src/DataTablesServiceProvider.php
+++ b/src/DataTablesServiceProvider.php
@@ -26,6 +26,7 @@ class DataTablesServiceProvider extends ServiceProvider
         $this->app->alias('datatables', DataTables::class);
         $this->app->singleton('datatables', fn () => new DataTables);
 
+        $this->app->alias('datatables.request', Request::class);
         $this->app->singleton('datatables.request', fn () => new Request);
 
         $this->app->singleton('datatables.config', DataTablesConfig::class);
--- a/src/DataTablesServiceProvider.php
+++ b/src/DataTablesServiceProvider.php
@@ -26,8 +26,11 @@ class DataTablesServiceProvider extends ServiceProvider
         $this->app->alias('datatables', DataTables::class);
         $this->app->singleton('datatables', fn () => new DataTables);
 
+        // Scoped and not shared, so that the request state, e.g. the flag set
+        // by ignoreMaxLength(), does not leak into the next request of a long
+        // running worker such as Octane.
         $this->app->alias('datatables.request', Request::class);
-        $this->app->singleton('datatables.request', fn () => new Request);
+        $this->app->scoped('datatables.request', fn () => new Request);
 
         $this->app->singleton('datatables.config', DataTablesConfig::class);
     }
--- a/src/Utilities/Request.php
+++ b/src/Utilities/Request.php
@@ -11,6 +11,11 @@ use Illuminate\Support\Facades\Config;
 class Request
 {
     /**
+     * Flag to ignore the configured maximum length.
+     */
+    protected bool $ignoreMaxLength = false;
+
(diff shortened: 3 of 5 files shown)


yajra:
Will have to check this against https://github.com/yajra/laravel-datatables-buttons. On the export process, we forced the length to `-1` to be able to download all filtered records. Does this change will affect that behavior?

https://github.com/yajra/laravel-datatables-buttons/blob/master/src/Services/DataTable.php#L344-L347

Thanks!

OzanKurt:
Good catch, thanks. I checked `getAjaxResponseData()` in buttons and yes, this would have truncated every export to `max_length`, since it forces `length => -1` to get all filtered records.

I've pushed an opt-out rather than special-casing `-1`, since anything derived from request input could be spoofed and would defeat the purpose of the cap:

```php
app('datatables.request')->ignoreMaxLength();
```

`datatables.request` is a singleton, so calling it in `getAjaxResponseData()` right before `app()->call($this->ajax(...))` covers the export. There's also a fluent `$dataTable->ignoreMaxLength()` for application code. Both are covered by tests, one of which mirrors the buttons export flow.

yajra (approved):

